PR #1, opened 2016-10-12
Example page

karelhala:
Let's kick off example application with npm and typscript.

Some example application is present so we can discuss styling and such.

If you want to build it locally you need to have npm installed locally, then run `npm run-script build` application will be present in `dist/` folder.

If you want to see changes, run `npm start` which will run node-dev server in memory (no changes present on disk).

lumeer


marvec:
Cool, many thanks! I used that yesterday, now merging...

karelhala:
Perfect, no problem. We can discuss rhe technologies and such over slack, some kind of brain storming and propositions on what to use might be good.

I hope it was welcomed good.

PR #4, opened 2016-11-16
Navigation

karelhala:
New routes are defined for each part of application. If we want to add new navigation items just add new items to router with `data` property with `title` set, same as in `src/components/forms/forms.module.ts`.

When navigating change header and breadcrumbs as well. Breadcrumbs will be clickable in future.

A lot of libraries were moved from node_packages to CDN.

selection_091
selection_092
selection_093


karelhala:
Just to be clear about loading npm packages via CDN not in vendor file.

Vendor file was almost 5MB large before this PR, now it's bellow 2MB so quite improvement, however I am planning on removing all vendor dependencies from this file and loading them from CDNs.

karelhala:
@marvec ping

PR #45, opened 2017-03-27
New navigation

karelhala:
As I was speaking with Martin it turns out that our navigation was quite hard to understand, so we came up with this new way.

### Old navigation to research
old_navigation_1

### New navigation to research
This is starting point for whole application, later it will be replaced with filter over all items (including views, collections and links)
new_navigation_1

### Old navigation to pick item
Here was small bug where it forgot filtered items
old_navigation_2

### New navigation to pick item
It will remeber filtered items now
new_navigation_2


karelhala:
@marvec @alicak @livthomas @kubedo8 ping.

No filter over all items, this will be step number two.

PR #46, opened 2017-03-27
Update npm dependencies to newest version

karelhala:
Update all npm dependencies to use newest versions, angular is now using `<ng-template></ng-template>` instead of `<template></template>`, so rather use this.

Also as [blog](https://angularjs.blogspot.cz/2017/03/angular-400-now-available.html) mentioned, angular moved animations to special module so import animations from them and not from core.


src/main/scripts/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "removeComments": false,
+    "skipLibCheck": true,
```
livthomas: Why do we want to do this?

src/main/scripts/src/components/views/research/document-preview.component.ts
```diff
@@ -1,9 +1,9 @@
-import {Component, trigger, state, style, transition, animate, keyframes, Input} from '@angular/core';
+import {Component, Input} from '@angular/core';
 import * as _ from 'lodash';
 import {LocalStorage} from 'ng2-webstorage';
 import {DocumentService} from '../../../services/document.service';
-import {DocumentInfoService} from '../../../services/document-info.service';
 import {QueryTagService} from '../../../services/query-tags.service';
+import { trigger, state, style, transition, animate, keyframes,} from '@angular/animations';
```
livthomas: There is unnecessary comma but I assume it is not a problem.

src/main/scripts/tsconfig.json
karelhala: If typescripts finds some kind of error in libs (`node_modules`) it won't throw any error. This is useful so the webpack build won't fail due to these errors (which we can't handle anyway)

karelhala:
I have added minification and AOT to build. Not sure if AOT is running at all, since nothing else just some webpack configuration has been done.

However final application is almost 1/3 of size with minification on (even when enabling gzip):

|Build type | vendor.js  | app.js | together|
| ------------- | ------------- | ------------- | ------------- |
|No minification | 3.4 MB  | 586kB  | 3.986 MB|
|With minification | 901 kB | 335kB  | 1.236 MB|
|No minification (gzipped)| 544 kB  | 90kB | 634 kB |
|With minification (gzipped)| 215 kB | 63 kB  | 278 kB|

karelhala:
@marvec @livthomas can we merge this?

PR #49, opened 2017-05-06
Update Angular Perfect Scrollbar dependency

livthomas:
There is a warning when installing the project with `angular2-perfect-scrollbar` so I have updated it to the latest version `ngx-perfect-scrollbar` which supports Angular 4.

The same problem is with `angular2-drag-scroll`, however, I have not been able to find a new version of this library for Angular 4.


karelhala:
Code looks good, I'll try locally.

PR #50, opened 2017-05-22
Provided basic connection with backend #LMR-157

kubedo8:


karelhala:
@kubedo8 from first glance it looks good! I'll check it more deeply tomorrow, however you should post some screenshots so we know how it looks like (I know that it connects BE to FE, but would be nice to see some actual data).

livthomas:
`company-project.service.ts` should be renamed to `organization-project.service.ts` to follow Angular conventions.

karelhala:
I think we can merge this, looks good and works good.

PR #53, opened 2017-05-30
Create new application layout #LMR-201

livthomas:
I have created a new application layout as suggested in the videos created by @marvec. So far it contains only the top panel but we will add more components in the next few weeks. In addition to the visual changes the whole structure of the application has been changed in order to prepare it for the future growth.

screencapture-localhost-7000-ui-workspace-1496178743580

I have not deleted any old files since @kubedo8 is going to move some functionality to this new structure. The application can still be easily switched to the old layout by reverting the change in `boot.ts` and `template-index.ejs` files.


karelhala:
Looks good to me, good to merge.

PR #54, opened 2017-06-06
Create/edit/delete orgorganizations and projects

kubedo8:
Firstly we need discuss this with @livthomas and edit some things..


src/main/scripts/src/app/shared/organization.ts
```diff
@@ -0,0 +1,13 @@
+import {Project} from './project';
+
+export class Organization {
```
livthomas: I would suggest moving this class and `Project` class to `app/shared/dto` folder since there will be many similar classes implemented in the future.

src/main/scripts/src/app/core/http-json.service.ts
```diff
@@ -18,13 +18,26 @@
...
+import {Http, Headers, RequestOptions} from '@angular/http';
 
-@Component({
-  selector: 'workspace',
-  template: require('./workspace.component.html'),
-  styles: [require('./workspace.component.scss').toString()]
-})
-export class WorkspaceComponent {
+@Injectable()
+export class HttpJson {
```
livthomas: I would suggest the following changes:

1. rename this class to `HttpClient`
2. implement all HTTP methods here
3. use this class everywhere instead of Angular's `Http` class

src/main/scripts/src/app/app.module.ts
```diff
@@ -30,6 +32,7 @@ import {SettingsModule} from './settings/settings.module';
     AppRoutingModule,
     BrowserModule,
     CoreModule,
+    HttpModule,
```
livthomas: What about importing this just in the *core* module?

src/main/scripts/src/app/settings/workspace-chooser/workspace-chooser.service.ts
```diff
@@ -0,0 +1,46 @@
...
+import {Injectable} from '@angular/core';
+import {Http} from '@angular/http';
+
+import 'rxjs/add/operator/map';
+import {Observable} from 'rxjs/Observable';
+import {Organization} from '../../shared/organization';
+import {Project} from '../../shared/project';
+
+@Injectable()
+export class WorkspaceChooserService {
```
livthomas: This is not needed. The methods should be moved to `OrganizationService` and `ProjectService`.

livthomas:
I think we need to store the last chosen organization and project for a given user on the server side as well. So you do not need to select it again every time you open our application.

src/main/scripts/src/app/settings/organization/organization.component.ts
```diff
@@ -53,20 +53,33 @@ export class OrganizationComponent implements OnInit {
 
   public onSave() {
     if (this.orgCode) {
-      this.organizationService.editOrganization(this.orgCode, this.organization);
+      this.organizationService.editOrganization(this.orgCode, this.organization)
+        .subscribe(response => {
+          if (response.ok) {
+            this.location.back();
```
livthomas: What if this operation fails?

karelhala: There are 2 ways of creating services which fetch data from server:

1. Inside observable create function which will fetch data, subscribes and assignes data to variable. This is prefect for example for data which doesn't need to be refreshed that much and can be shared amongst other components
2. Create observable and subscribe to it in components. This is prefered when data should not be shared in components and whenever component is presented and rendered it will fetch newest data from server.

First solution has advantage of sharing data, but the data might be outdated (we have to refresh the data). Second solution avoids problem with refreshing data when rendering, but if we'd need to refresh data which are already fetched in component there is no way of triggering this refresh.

PR #55, opened 2017-07-02
Profile dropdown

justmesr:
Profile dropdown-menu with placeholder links for Settings and user licence type.

### DIFF
--- a/src/main/scripts/src/app/core/header/header.component.html
+++ b/src/main/scripts/src/app/core/header/header.component.html
@@ -22,8 +22,8 @@
   <a #profile class="profile">
 
     <div *ngIf="optionsVisible" class="arrow-up"></div>
-    <ul #options *ngIf="optionsVisible" [@animateHeight]="'in'">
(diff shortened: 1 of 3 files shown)


src/main/scripts/src/app/core/header/header.component.html
```diff
@@ -17,5 +17,17 @@
...
   </div>
   <div class="notifications">0</div>
-  <div class="profile"></div>
+
+  <!-- User icon -->
+  <a #profile class="profile">
+
+    <div *ngIf="optionsVisible" class="arrow-up"></div>
+    <ul #options *ngIf="optionsVisible" [@animateHeight]="'in'">
+      <li><a href=""><i class="fa fa-cog" aria-hidden="true"></i>Settings</a></li>
```
livthomas: This link should point to `/user/settings`. See the examples of links in [other parts of the application](https://github.com/Lumeer/web-ui/blob/devel/src/main/scripts/src/app/core/header/header.component.html#L15).

PR #56, opened 2017-07-02
Fix workspace service issues #LMR-230

livthomas:


src/main/scripts/src/app/core/workspace.service.ts
```diff
@@ -30,4 +30,10 @@ export class WorkspaceService {
   public organizationCode: string;
   public projectCode: string;
 
+  public isWorkspaceSet(): boolean {
+    this.organizationCode = undefined;
```
karelhala: No need to assign undefined in `is`Function, if orgCode is not set (as seen above) default value is undefined.

Also no need to check length if variable is string shorter version of return should be

```javascript
return this.organizationCode && this.projectCode;
```

src/main/scripts/src/app/core/header/header.component.html
```diff
@@ -11,10 +11,10 @@
       <span>Search</span>
     </button>
   </div>
-  <div class="project">
+  <div class="project" *ngIf="workspaceService.isWorkspaceSet()">
     <a routeLink="/workspace">
```
karelhala: I guess based on Martin's comments about apache reverse proxy we should add href to `a` tag

```HTML
<a href="javascript: void(0)" routerLink="/workspace">...</a>
```

src/main/scripts/src/app/core/workspace.service.ts
livthomas: Sorry, this was just a test I forgot to remove. When it comes to that condition, I have tried the one you suggest but it does not work in Typescript.

src/main/scripts/src/app/core/header/header.component.html
livthomas: What is this good for?

PR #59, opened 2017-07-07
Configure advanced application routing #LMR-232

livthomas:
- Use multiple routing modules.
- Introduce Not Found page.
- Rename Settings module and its components.
- Create Collection module with routing skeleton.
- Move REST services and DTOs to core module.


livthomas:
Route guards will be introduced in the future pull-requests.

PR #62, opened 2017-07-11
Draft perspectives structure and usage #LMR-214

livthomas:
You can try it by accessing the following address:

http://localhost:7000/ui/w/LMR/devel/search/documents?perspective=Table&query=something

### DIFF
--- a/src/main/scripts/src/app/search/documents/documents-search.component.ts
+++ b/src/main/scripts/src/app/search/documents/documents-search.component.ts
@@ -1,0 +1,66 @@
+/*
+ * -----------------------------------------------------------------------\
+ * Lumeer
+ *
+ * Copyright (C) since 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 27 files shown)


src/main/scripts/src/app/search/documents/documents-search.component.ts
```diff
@@ -0,0 +1,78 @@
...
+    let componentFactory = this.componentFactoryResolver.resolveComponentFactory(perspectiveComponent);
+
+    let viewContainerRef = this.perspectiveDirective.viewContainerRef;
+    viewContainerRef.clear();
+
+    let componentRef = viewContainerRef.createComponent(componentFactory);
+    (<Perspective>componentRef.instance).query = this.query;
+  }
+
+  private static getDocumentsPerspectiveComponent(perspective: DocumentsPerspective): Type<any> {
```
karelhala: You don't have to use switch (which is kinda slow), but you can hash table with types. Does not matter where you will create it, either in util class, or in interface definition.
```Typescript
const perspectives = {
    [DocumentsPerspective.Table]: TableDocumentsPerspectiveComponent,
    [DocumentsPerspective.PostIt]: PostItDocumentsPerspectiveComponent
};
```
```Typescript
return perspectives[perspective] || PostItDocumentsPerspectiveComponent;
```

PR #66, opened 2017-07-15
Fix profile dropdown, header elements squish, webpack context warning

justmesr:


src/main/scripts/src/app/core/header/header.component.scss
```diff
@@ -97,6 +102,7 @@
 
   ul {
     box-shadow: $default-shadow;
+    background: white;
```
karelhala: Please use `$secondary-color` - which is translated to `#ffffff` instead of just `white` color.

src/main/scripts/src/app/core/header/header.component.scss
```diff
@@ -69,6 +71,9 @@
 }
 
 .profile {
+  min-width: 45px;
+  min-height: 45px;
```
livthomas: The following code does the same thing in our layout:
```css
flex-shrink: 0;
```
But your solution is valid as well.

PR #67, opened 2017-07-18
Add search box with suggestions #LMR-179

livthomas:
I have improved the search box to be able to suggest as well as build queries.

### DIFF
--- a/src/main/scripts/src/app/core/header/header.component.html
+++ b/src/main/scripts/src/app/core/header/header.component.html
@@ -4,30 +4,28 @@
       <span>Lumeer</span>
     </a>
   </div>
-  <!-- TODO search box will be provided by another component -->
   <div class="search">
-    <input class="search-box" placeholder="Type anything you search for..."/>
-    <button class="search-button">
-      <span>Search</span>
-    </button>
+    <search-box *ngIf="isSearchBoxShown()"></search-box>
   </div>
-  <div class="project" *ngIf="workspaceService.isWorkspaceSet()">
-    <a href="javascript: void(0)" routeLink="/workspace">
-      <span>{{workspaceService.organizationCode}}@{{workspaceService.projectCode}}</span>
+  <div class="menu">
+    <div class="project" *ngIf="workspaceService.isWorkspaceSet()">
+      <a routeLink="/workspace">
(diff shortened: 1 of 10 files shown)


src/main/scripts/src/app/core/header/header.component.html
```diff
@@ -4,16 +4,12 @@
       <span>Lumeer</span>
     </a>
   </div>
-  <!-- TODO search box will be provided by another component -->
   <div class="search">
-    <input class="search-box" placeholder="Type anything you search for..."/>
-    <button class="search-button">
-      <span>Search</span>
-    </button>
+    <search-box *ngIf="workspaceService.isWorkspaceSet()"></search-box>
```
kubedo8: Isn't property `showSearchBox` planned for this situation?

src/main/scripts/src/app/core/rest/search.service.ts
```diff
@@ -0,0 +1,48 @@
...
+@Injectable()
+export class SearchService {
+
+  constructor(private http: Http, private workspaceService: WorkspaceService) {
+  }
+
+  public suggest(text: string, type: string): Observable<Suggestion[]> {
+    return this.http
+      .get(`${this.searchPath()}/suggestion?text=${text}&type=${type}`)
+      .map(response => response.json() as Suggestion[]);
```
kubedo8: There is better way to add query parameters, but we will use new `HttpClient` instead and change it...

kubedo8:
Can you include some picture to demonstrate new functionality?

src/main/scripts/src/app/core/rest/search.service.ts
livthomas: Thanks, I haven't even used our `HttpClient` here. I will refactor this using a new Angular client.

src/main/scripts/src/app/core/header/header.component.html
livthomas: Good catch!

livthomas:
screencapture-localhost-7000-ui-w-lmr-devel-search-1500465554759

livthomas:
It looks like this now:
screencapture-localhost-7000-ui-w-lmr-devel-search-1500541555099

PR #69, opened 2017-07-18
Correct workspace set

kubedo8:


src/main/scripts/src/app/app.component.ts
```diff
@@ -18,14 +18,36 @@
...
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
+
+  constructor(private workspaceService: WorkspaceService,
+              private router: Router) {
+  }
+
+  public ngOnInit(): void {
+    this.router.events.subscribe(val => {
```
livthomas: I would call this variable `event` instead of `val` since it is not obvious what it is at first sight.

src/main/scripts/src/app/shared/perspectives/documents/documents-perspective-presenter.ts
```diff
@@ -31,7 +31,7 @@ export abstract class DocumentsPerspectivePresenter extends PerspectivePresenter
     super(activatedRoute, componentFactoryResolver);
   }
 
-  public ngAfterViewInit() {
+  public ngOnInit() {
     this.loadPerspectiveComponent(DocumentsPerspective.getPerspective(this.perspective).component);
```
livthomas: This is not a good idea since this overrides parent `ngOnInit()` method. We should rather specify abstract method `loadPerspective()` in `PerspectivePresenter` and implement it here instead of this `ngOnInit()`. This method should also be called at the end of `ngOnInit()` method in `PerspectivePresenter`.

PR #70, opened 2017-07-21
Create collections module #LMR-196

kubedo8:


kubedo8:
coll1

@karelhala I can't make icons scrollable horizontal (as it was before). Could you help me or do you fix it later?
coll2

src/main/scripts/src/app/core/local.service.ts
```diff
@@ -0,0 +1,42 @@
...
+ * -----------------------------------------------------------------------/
+ */
+
+import {Injectable} from '@angular/core';
+
+import {Observable} from 'rxjs/Observable';
+import {HttpClient} from './rest/http-client.service';
+
+@Injectable()
+export class LocalService {
```
livthomas: What is this service good for? I would suggest moving both JSONs to typescript constants.

src/main/scripts/src/app/app.component.ts
```diff
@@ -18,14 +18,36 @@
...
+import {WorkspaceService} from './core/workspace.service';
 
 @Component({
   selector: 'app',
   templateUrl: './app.component.html',
   styleUrls: ['../styles/basic.scss', './app.component.scss'],
   encapsulation: ViewEncapsulation.None
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
```
livthomas: These changes have already been made in the other pull request.

src/main/scripts/src/app/core/rest/collection.service.ts
```diff
@@ -19,27 +19,34 @@
...
 export class CollectionService {
 
   constructor(private http: HttpClient, private workspaceService: WorkspaceService) {
   }
 
-  public getCollections(pageNumber?: number, pageSize?: number): Observable<any[]> {
-    let queryParams = !isUndefined(pageNumber) && !isUndefined(pageSize) ? `?page=${pageNumber}&size=${pageSize}` : '';
+  public getCollections(pageNumber?: number, pageSize?: number): Observable<Collection[]> {
+    let queryParams;
+    if (pageNumber && pageSize) {
```
livthomas: What happens if `pageNumber` is `0`? We should rather use `isUndefined` method or something similar here.

src/main/scripts/src/app/core/rest/collection.service.ts
```diff
@@ -19,27 +19,34 @@
...
+      let queryParams = new HttpParams().set('page', pageNumber.toString())
+        .set('size', pageSize.toString());
+    }
+
+    return this.http.get<Collection[]>(this.apiPrefix(), queryParams ? {params: queryParams} : undefined);
```
livthomas: Why this check? Cannot we just pass empty query parameters?

src/main/scripts/src/app/core/rest/collection.service.ts
```diff
@@ -19,27 +19,34 @@
...
 @Injectable()
 export class CollectionService {
 
   constructor(private http: HttpClient, private workspaceService: WorkspaceService) {
   }
 
-  public getCollections(pageNumber?: number, pageSize?: number): Observable<any[]> {
-    let queryParams = !isUndefined(pageNumber) && !isUndefined(pageSize) ? `?page=${pageNumber}&size=${pageSize}` : '';
+  public getCollections(pageNumber?: number, pageSize?: number): Observable<Collection[]> {
+    let queryParams;
```
livthomas: You can assign `new HttpParams()` value here and use builder methods on `queryParams` in the block below.

src/main/scripts/src/app/core/rest/document.service.ts
```diff
@@ -31,9 +30,14 @@ export class DocumentService {
   }
 
   public getDocuments(collectionName: string, pageNumber?: number, pageSize?: number): Observable<any[]> {
-    let queryParams = !isUndefined(pageNumber) && !isUndefined(pageSize) ? `?page=${pageNumber}&size=${pageSize}` : '';
-
-    return this.http.get(this.apiPrefix(collectionName) + queryParams)
+    let queryParams;
+    if (pageNumber && pageSize) {
+      let queryParams = new URLSearchParams();
```
livthomas: The same comments as in `CollectionService` apply here as well.

src/main/scripts/src/app/core/rest/document.service.ts
```diff
@@ -22,7 +22,6 @@ import {Injectable} from '@angular/core';
 import {HttpClient} from './http-client.service';
```
livthomas: Cannot we use a new Angular HTTP client directly?

src/main/scripts/src/app/shared/perspectives/collections/post-it/post-it-collections-perspective.component.ts
```diff
@@ -18,23 +18,87 @@
...
+      this.colors = colors;
+    });
+  }
+
+  public saveCollection() {
+    this.collectionService.saveCollection(this.newCollection)
+      .subscribe((code: String) => {
+        this.newCollection = new Collection();
+        this.collectionService.getCollections()
+          .subscribe((collections: Collection[]) => {
```
livthomas: Why do you subscribe again? It has already been done in the `ngOnInit()` method so it should react to changes.

src/main/scripts/src/app/shared/perspectives/documents/documents-perspective.ts
```diff
@@ -37,11 +36,11 @@ export class DocumentsPerspective {
   public static defaultPerspective = DocumentsPerspective.PostIt;
 
   public static getPerspective(perspectiveId: string): DocumentsPerspective {
-    if (isNullOrUndefined(perspectiveId)) {
+    if (!perspectiveId) {
```
livthomas: Why these changes?

marvec:
It was the original request to have colors and icons together to limit clicking count. The icons should be grouped like emoticons on a mobile phone. There are too many icons to show them all at once. When a new record appears next to the plus one, it should have name input focused immediately.

marvec:
And we need to have many more colors. Like the palette used for text color selection in Google Docs.

livthomas:
We have discussed with @marvec that the color and icon picker should be shown when you click on an icon of any collection.

Note that this picker should be moved to a separate component in `app/shared/perspectives/shared/icon-picker` folder so it can be reused by multiple perspectives.

src/main/scripts/src/app/core/local.service.ts
karelhala: It depends if we want to use only these two sets of colors and icon. We could store the icons and colors on server in some kind of smart way (most used icons for current user...).

src/main/scripts/src/app/shared/perspectives/collections/post-it/post-it-collections-perspective.component.ts
karelhala: Subscribe does not work like that. Whenever you subscribe to service, the service is fired and function is executed. If we want to have lazy loaded collections we have 2 options (angular's way, or Rx.js way), both are valid.
1) Angular's way:
We'd have to do something like this ` this.collections = this.collectionService.getCollections()` and later on in template (notice the `async` pipe in `*ngFor`).
```html
<div class="col-sm-3 post-it" *ngFor="let oneCollection of collections | async">
  <span>Documents: {{oneCollection.documentCount}}</span>
  <div class="info">
    <i class="fa fa-2x" [ngClass]="oneCollection.icon" [ngStyle]="{'color':oneCollection.color}"></i>
    <h4>{{oneCollection.name}}</h4>
  </div>
</div>
```

2) Rx.js way
Using [behaviorsubject](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/subjects/behaviorsubject.md) which reacts to every change of promis, example can be seen at [behaviorsubject at scotch](https://scotch.io/tutorials/3-ways-to-pass-async-data-to-angular-2-child-components#Scotch.io_970x90_728x90_320x50_ATF_InPage3)

src/main/scripts/src/app/shared/perspectives/collections/post-it/post-it-collections-perspective.component.ts
```diff
@@ -18,23 +18,87 @@
...
 
   @Input()
   public query: string;
 
   @Input()
   public editable: boolean;
 
-  constructor(private collectionService: CollectionService) {
+  public placeholderTitle: string = 'Collection name';
+  public iconsPerPage: number = 36;
```
karelhala: Can this be stored in some interface as constant? So we can change it easilly and we have exact place where to go to see which constants are used.

karelhala:
I would recommend merging the color and icon picker as it is (based on the look and feel), we can improve it later on.

Also based on the screenshot @kubedo8 add small icon with selected color to each tile. User would be able to click on it and change both icon and color.

About more colors, what about we'd show couple of most used colors for each user (fetched from server) and button which would show extended color picker, maybe [html input color](https://www.w3schools.com/html/tryit.asp?filename=tryhtml_input_color) or [fancy color picker](https://github.com/Alberplz/angular2-color-picker).

About grouped icons 👍 on it, but I would improve it later down the line, so now we can add some basic items with color and icons.

src/main/scripts/src/app/core/rest/document.service.ts
kubedo8: That's out of my scope.

kubedo8:
After some changes collection perspective looks like this. I think that adding new collection is not intuitive, so we can discuss how to improve it...
Color & Icon picker stays the same, because @justmesr already implemented it, so I will wait for future commit...

col3

src/main/scripts/src/app/app.component.scss
```diff
@@ -11,7 +11,7 @@ header {
 }
 
 main {
-  width: 100vw;
-  padding: 0 5vw;
-  overflow-y: auto;
+  margin: 10px;
+  height: calc(100vh - 90px);
```
livthomas: I am not sure about this. Search box can grow vertically and so the header might be higher than `90px`.

src/main/scripts/src/app/shared/const.ts
```diff
@@ -0,0 +1,80 @@
+export const placeHolderNewCollection = 'Collection name';
```
livthomas: I think this does not need to be here. It will be one of the localized strings in the feature and almost certainly put somewhere else. So for now, I think can stay directly inside the component where it is needed.

kubedo8:
I changed way how to add new collection as we discussed with @livthomas.
- Collection is added after losing focus on input box (and name length is gte 3)
- Collection is updated after losing focus on input box  and also if icon and color are setted.
- Collection has gray background only if is not already in database, and there's option to delete it (locally)

ezgif-1-ed1b9ddf87

src/main/scripts/src/app/app.component.scss
kubedo8: But this is the only way I was able to make main content scrollable without scrolling header. Do you know better solution?

src/main/scripts/src/app/shared/const.ts
kubedo8: Ok..

src/main/scripts/src/app/app.component.scss
livthomas: We will probably need to change the whole base layout using Bootstrap in order to solve this. So keep it like this for now and we will fix it in the future.

kubedo8:
Added animations and shadow to input box with hover animation like google searchbox.
ezgif-1-92fc788d3e

PR #72, opened 2017-07-24
[WIP] Editable table

karelhala:
### Introduces basic `static-table`
This PR creates new component which is stored in shared module and is called `StaticTableComponent`, this component is starting point of more complex table, which will be core principle of Lumeer. No connections, no auto complete, etc. These functions will be added in later PRs.

Currently supported events for static table:
* [x] Highlight row and col of active item.
* [x] Header editable
* [x] Columns editable
* [x] Remove column
* [x] Remove row
* [x] Hide column
* [x] Hide row
* [x] Add new column
* [x] Add new row

peek 2017-07-24 15-43

### Updated look and feel
peek 2017-08-11 16-29


src/main/scripts/src/vendor.ts
```diff
@@ -23,4 +23,5 @@ import 'angular2-drag-scroll';
 // import 'lodash';
 import 'ngx-perfect-scrollbar';
 import 'ng2-webstorage';
+import 'angular2-virtual-scroll';
```
livthomas: You forgot to add `angular2-virtual-scroll` dependency as well.

src/main/scripts/src/app/shared/table/static-table/static-table.component.html
```diff
@@ -0,0 +1,25 @@
+<div>
```
livthomas: Why don't we use `<table>` here? And we should set `border-collapse: collapse` on it.

karelhala: Css tables are easier to change based on device, so the table can look slightly different on smaller device screen. Or it can has sticky footer, easier to position additional content (like double border in joined records) and such.

Thank you for pointing out the border-collapse :+1:

karelhala:
Future improvements (not recommended in this PR)
* Drag and drop on headers to reorder whole column
* Resize on header
* Mass select of items

And of course connection to different table, which can be hooked to either while typing header name, or when adding new column via plus button.

livthomas:
How will `static-table` communicate with a parent component? Do we just want to emit various events in case of changes (row/column added/removed, cell changed, etc.) ? Or do we plan to use two-way data binding?

PR #74, opened 2017-07-26
The implementation of an abstract permissions component #LMR-247

martinglova:


src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
   public entities: Entity[];
   public otherEntities: string[];
 
-  public constructor(private rolesService: RolesService) {
+  public rolesCheckbox;
```
livthomas: What is the type of this field?

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
   public ngOnInit() {
     this.roles = ROLES[this.resourceType];
-    this.entities = this.getEntities(this.entityType);
-    this.otherEntities = this.getOtherEntities(this.entityType);
+    this.rolesCheckbox = {};
+    this.roles.forEach(r => {
+      this.rolesCheckbox[r] = false;
+    });
+    switch (this.resourceType) {
+      case 'organization':
```
livthomas: Note that this component should work for collections and views as well.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
+  }
+
+  public updateCheckedRoles(role: string, event) {
+    this.rolesCheckbox[role] = event.target.checked;
+  }
+
+  public addOrganizationUsersGroupsRole(orgCode: string, role: string, users: string[], groups: string[]) {
+    this.rolesService.addOrganizationUsersGroupsRole(orgCode, role, users, groups)
+      .subscribe(
+      response => {;}
```
livthomas: This is not needed. You can call just `.subscribe()`.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
+    let addedRoles = [];
+    switch (this.resourceType) {
+      case 'organization':
+        this.route.paramMap.subscribe((params: ParamMap) => {
+          let orgCode: string = params.get('organizationCode');
+          switch (this.entityType) {
+            case 'users':
+              for (let r in this.roles) {
+                if (this.rolesCheckbox[this.roles[r]] && this.roles.hasOwnProperty(r)) {
+                  this.addOrganizationUsersGroupsRole(orgCode, this.roles[r], [name], []);
```
livthomas: This is against clean code principles. We should never have more than 3 levels of nesting! Try to extract some parts of this method to new methods such as `addOrganizationRole()` and `addOrganizationRoleToUsers()` inside it, etc.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
+                  addedRoles.push(this.roles[r]);
+                }
+              }
+            break;
+            case 'groups':
+              for (let r in this.roles) {
+                if (this.rolesCheckbox[this.roles[r]] && this.roles.hasOwnProperty(r)) {
+                  this.addOrganizationUsersGroupsRole(orgCode, this.roles[r], [], [name]);
+                  addedRoles.push(this.roles[r]);
```
livthomas: This is exactly the same block of code as in previous case. Try to extract it to a single generic function with two optional parameters.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
+              break;
+          }
+        });
+    }
+    this.otherEntities.push(this.entities[index]['name']);
+    this.entities.splice(index, 1);
+  }
+
+  public changePermission(entityName: string, role: string, event) {
+    let addPermision: boolean = event.target.checked;
```
livthomas: You do not need to put type annotations on variables if you assign values directly. In such cases types are automatically inferred by TypeScript.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
+                  this.removeProjectUsersGroupsRole(orgCode, proCode, role, [], [entityName]);
+              }
+          }
+          break;
+      }
+    });
+  }
+
+  private getEntitiesForUserOrGroup(userOrGroup: string, roles: Role[]): Entity[] {
+    let arr = [];
```
livthomas: What does this variable represent? Try to use more self-explaining names even though they are longer.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,308 @@ export class PermissionsTableComponent implements OnInit {
...
+    arr.push({});
+    roles.forEach(e => {
+      e[userOrGroup].forEach(u => {
```
livthomas: Try to avoid using one-letter variable names. The code is less readable.

src/main/scripts/src/app/core/dto/users.ts
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------/
+ */
+
+ export class Users {
+
+     public map: Map<string, string[]>;
```
livthomas: Why is there a `map` field? It is not obvious what exactly this class represents and what kind of data it keeps.

Also use proper code formatting according to our code style.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.html
```diff
@@ -7,27 +7,33 @@
...
-        <input type="checkbox" title="{{role}}" [attr.checked]="entity.roles.indexOf(role) > -1 ? '': null"/>
+      <td *ngFor="let role of roles"  class="role">
+        <div class="material-switch pull-left">
+          <input id="ch{{entity.name+role}}" type="checkbox" title="{{role}}" [attr.checked]="entity.roles.indexOf(role) > -1 ? '': null" (change)="changePermission(entity.name, role, $event)"/>
+          <label for="ch{{entity.name+role}}" class="label-primary"></label>
+        </div>
       </td>
       <td>
-        <button (click)="onRemove(entity.name)" class="btn btn-remove">Remove</button>
+        <button (click)="onRemove(i)" class="btn btn-danger">Remove</button>
```
livthomas: Why cannot we use entity name here?

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
martinglova: Adding new cases will provide that. But I wanted to implement it once CRUD operations for collections and views are implemented in UI.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
martinglova: It is not the same. Check the order of the parameters of the `addOrganizationUsersGroupsRole` function. In the first case, a role is added for a user in the second for a group.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.html
martinglova: The `entities` variable keeps the array of entity objects. When I want to remove an item of an array I need the index. This way I do not have to get the index of the element I want to remove in the `onRemove` method.
If there is only the name of an entity in `onRemove` method (as it was before) I need to iterate the whole `entities` array and look for the entity object with the given name to get the index.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
martinglova: I changed it, but I am not very sure if the '3 levels of nesting' rule is always strictly the best solution. Sometimes if you do something in code really once and the level of nesting is slightly higher you need to scroll a lot to keep track what is happening in the code. I think, in such cases, a higher level of nesting would be better.

kubedo8:
You added also my commit, so you should remove it.

src/main/webapp/WEB-INF/keycloak.json
```diff
@@ -6,5 +6,6 @@
   "public-client": true,
   "disable-trust-manager": "false",
   "truststore": "${jboss.server.config.dir}/cacerts.jks",
-  "truststore-password": "geheim"
+  "truststore-password": "geheim",
+  "disabled": true
```
livthomas: You should not commit this!

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -47,54 +52,342 @@ export class PermissionsTableComponent implements OnInit {
...
+              this.addProjectUsersGroupsRole(orgCode, proCode, this.roles[roleIdx], [name], []);
+              addedRoles.push(this.roles[roleIdx]);
+            }
+          }
+        break;
+        case 'groups':
+          for (let roleIdx in this.roles) {
+            if (this.rolesCheckbox[this.roles[roleIdx]] && this.roles.hasOwnProperty(roleIdx)) {
+              this.addProjectUsersGroupsRole(orgCode, proCode, this.roles[roleIdx], [], [name]);
+              addedRoles.push(this.roles[roleIdx]);
```
livthomas: The same block of code is repeated four times above. It can be extracted into one new method if properly parametrized.

PR #77, opened 2017-08-04
Fix search button icon #LMR-289

SquattingPenguin:


src/main/scripts/src/app/shared/search-box/search-box.component.html
```diff
@@ -12,6 +12,6 @@
   </div>
 </div>
 
-<button (click)="onButtonClick()">
+<button class="fa fa-search fa-fw" (click)="onButtonClick()">
```
livthomas: Why not using the default font-awesome way of adding icons to your code?
```html
<i class="fa fa-search" aria-hidden="true"></i>
```
You can add this inside the `<button>` element just before `<span>`.

SquattingPenguin:
It looks like this now : 
tom1
tom2

src/main/scripts/src/app/shared/search-box/search-box.component.html
justmesr: Although it's basically the same, isn't it better the way it is? The way I see the difference between the two options is: 
- The button is the magnifying glass.
- The button contains the magnifying glass.

PR #82, opened 2017-08-16
Env variables

karelhala:
This will enable to parametrize build so we can pass environment variables to our application.

### LUMEER_ENV
To switch between different types of build process. LUMEER_ENV is always set to development for local changes (no need to change it locally). Also, if this variable is not set production mode is enabled.

### LUMEER_ENGINE
To set engine URL prefix, default value is `lumeer-engine`.

### DIFF
--- a/src/main/scripts/package.json
+++ b/src/main/scripts/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "webpack -p",
-    "start": "webpack-dev-server",
+    "start": "cross-env LUMEER_ENV=development webpack-dev-server",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "repository": {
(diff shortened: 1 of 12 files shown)


karelhala:
@marvec to change url under which server asks for REST data you have to set env variable `LUMEER_ENGINE` to something of your choice.

I did not went with HTTP interceptor, because it would require much more work and this way it's somewhat simpler.

marvec:
What are the valid values of LUMEER_ENV? LUMEER_ENGINE should not contain any slashes?

karelhala:
@marvec `LUMEER_ENV` currently we have only `development` supported because based on this variable we switch between using keycloak and not using keycloak. `LUMEER_ENGINE` is just string without slashes at start and end. further we can improve this by adding HTTP interceptor, check if this cariable contains slash at start and end and such things. But for now I feel this is fastest and cleanest way to do this.

@livthomas good point. I'll update this in follow up PR.

marvec:
So my question still is, how do I effectively set the production environment. LUMEER_ENV accepts only development, which is the default value. If I do not set the env. var, I get development. No other value is acceptable. So how to switch to production?

karelhala:
@marvec default value is production https://github.com/Lumeer/web-ui/pull/82/files#diff-7e99bc0d9bd3870289658153781316c8L5 if you do not set any value to `LUMEER_ENV` you will have production set up. However it has no effect, yet (except when running development keycloak is disabled)

PR #83, opened 2017-08-16
Restructure workspace module #LMR-297

SquattingPenguin:
ws


src/main/scripts/src/app/workspace/organization/form/organization-form.component.ts
```diff
@@ -44,17 +44,31 @@ export class OrganizationEditComponent implements OnInit {
...
       this.organizationCode = params.get('organizationCode');
       if (this.organizationCode) {
-        this.organizationService.getOrganization(this.organizationCode)
-          .subscribe((organization: Organization) => this.organization = organization,
-            error => this.errorMessage = error
-          );
+        this.getOrganization();
+      } else {
+        if (this.route.parent != null) {
+          this.route.parent.paramMap.subscribe((params: ParamMap) => {
```
livthomas: Why do we check both current route and the parent one? Don't we know in which context this component is used?

src/main/scripts/src/app/workspace/project/form/project-form.component.ts
```diff
@@ -49,15 +49,29 @@ export class ProjectEditComponent implements OnInit {
...
+            if (this.projectCode) {
+              this.getProject();
+            }
+          });
+        }
       }
     });
   }
 
+  public getProject(): void {
```
livthomas: This should also be `private`.

src/main/scripts/src/app/workspace/project/permissions/project-permissions.component.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  private project: Project;
+  private organizationCode: string;
+  private projectCode: string;
+  private errorMessage: any;
+
+  constructor(private projectService: ProjectService,
+              private workspaceService: WorkspaceService,
+              private route: ActivatedRoute,
+              private router: Router) {
```
livthomas: The same here. None of these is needed in this component.

src/main/scripts/src/app/workspace/workspace-routing.module.ts
```diff
@@ -21,15 +21,44 @@
...
+        path: 'settings',
+        component: OrganizationFormComponent
+      },
+      {
+        path: 'permissions',
+        component: OrganizationPermissionsComponent
+      },
+    ]
+  },
+  {path: 'organization/:organizationCode/project/add', component: ProjectFormComponent},
```
livthomas: This path should be before the organization path since it is more general.

src/main/scripts/src/app/workspace/workspace-routing.module.ts
```diff
@@ -21,15 +21,44 @@
...
+    children: [
+      {
+        path: 'settings',
+        component: OrganizationFormComponent
+      },
+      {
+        path: 'permissions',
+        component: OrganizationPermissionsComponent
+      },
+    ]
```
livthomas: There should full path redirect from `''` to `'settings'` here.

src/main/scripts/src/app/workspace/workspace-routing.module.ts
```diff
@@ -21,15 +21,44 @@
...
+    children: [
+      {
+        path: 'settings',
+        component: ProjectFormComponent
+      },
+      {
+        path: 'permissions',
+        component: ProjectPermissionsComponent
+      },
+    ]
```
livthomas: The same redirect should be added as the last entry here.

src/main/scripts/src/app/workspace/workspace.module.ts
```diff
@@ -34,13 +38,17 @@ import {SharedModule} from '../shared/shared.module';
...
+    ProjectPermissionsComponent,
+    OrganizationSettingsComponent,
+    ProjectSettingsComponent,
     WorkspaceChooserComponent
   ],
   exports: [
-    OrganizationEditComponent,
-    ProjectEditComponent,
+    OrganizationFormComponent,
+    ProjectFormComponent,
```
livthomas: I do not think these components need to be exported since they are not used in any other modules.

src/main/scripts/src/app/workspace/workspace-routing.module.ts
```diff
@@ -21,15 +21,64 @@
...
+      },
+      {
+        path: '',
+        redirectTo: 'settings',
+        pathMatch: 'full'
+      }
+    ]
+  },
+  {
+    path: 'organization/:organizationCode/project/add',
```
livthomas: You need to put this path before the previous one, otherwise it does not work because `add` is used as `projectCode` parameter.

src/main/scripts/src/app/workspace/workspace-routing.module.ts
```diff
@@ -21,15 +21,64 @@
...
+      {
+        path: '',
+        redirectTo: 'settings',
+        pathMatch: 'full'
+      }
+    ]
+  },
+  {
+    path: 'organization/add',
+    component: OrganizationFormComponent
```
livthomas: The same problem is here.

PR #84, opened 2017-08-18
Connect documents table with backend LMR-306

kubedo8:
Now it's possible to create document, update some value, add/remove/edit column through table perspective.

### DIFF
--- a/src/main/scripts/src/app/shared/perspectives/documents/table/table-documents-perspective.component.ts
+++ b/src/main/scripts/src/app/shared/perspectives/documents/table/table-documents-perspective.component.ts
@@ -35,6 +35,7 @@ import {TableHeaderCell} from '../../../table/model/table-header-cell';
 import {Attribute} from '../../../../core/dto/attribute';
 import {Document} from '../../../../core/dto/document';
 import {TableRowCell} from '../../../table/model/table-row-cell';
+import {DataEvent} from '../../../table/event/data-event';
 
 @Component({
   selector: 'table-documents-perspective',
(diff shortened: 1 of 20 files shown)


src/main/scripts/src/app/shared/table/static-table/static-table.component.ts
```diff
@@ -18,82 +18,139 @@
...
   }
 
   public onUpdateRow(rowIndex, dataPayload): void {
-    this.data.rows[rowIndex][dataPayload.index].label = dataPayload.data;
+    this.rows[rowIndex].cells[dataPayload.colIndex].label = dataPayload.data;
+    let header: string = this.header.cells[dataPayload.colIndex].label;
+    let value: any = {};
+    value[header] = dataPayload.data;
+    if (this.rows[rowIndex].id) {
+      value['id'] = this.rows[rowIndex].id;
```
livthomas: I think `id` should not be stored together with values.

src/main/scripts/src/app/shared/perspectives/documents/table/table-documents-perspective.component.ts
```diff
@@ -18,23 +18,135 @@
...
+      }
+    });
+    return document;
+  }
+
+  private prepareTableData(attributes: Attribute[], documents: Document[]) {
+    let headerRows: TableHeaderCell[] = attributes.map(attr =>
+      <TableHeaderCell>{label: attr.name, active: false, hidden: false, constraints: attr.constraints});
+    this.header = <TableHeader> {cells: headerRows};
+    this.rows = documents.map(document => {
```
livthomas: This method is too complicated and should be split into multiple simpler methods.

src/main/scripts/src/app/shared/table/model/table-row-cell.ts
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ * -----------------------------------------------------------------------/
+ */
+
+export interface TableRowCell {
+
+  label: string;
+  active: boolean;
+  hidden: boolean;
+  constraints: string[];
```
kubedo8: Because we have lines as separate component. So while editing value on some cell, we don't have access to header (which holds constraints).
Same with `hidden`, each cell has it's own css class.

PR #85, opened 2017-08-19
Enable ECMAScript 2017 constructs

livthomas:
@kubedo8, we will be able to use new functionality such as [Array.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) now.


karelhala:
LGTM

PR #87, opened 2017-08-20
Update to TypeScript 2.4

livthomas:
I also had to:
* Fix return types of service methods.
* Update to newer RxJS with fixed type problems.

We will be able to take full advantage of [new TypeScript features](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html) such as String Enums, etc.

### DIFF
--- a/src/main/scripts/package.json
+++ b/src/main/scripts/package.json
@@ -52,7 +52,7 @@
     "ts-loader": "2.0.3",
     "tslint": "^4.3.1",
     "tslint-loader": "^3.3.0",
-    "typescript": "~2.3.0",
+    "typescript": "~2.4.0",
     "typings": "2.1.0",
     "url-loader": "^0.5.7",
     "webpack": "^2.2.0",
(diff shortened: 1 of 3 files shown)


src/main/scripts/src/app/core/rest/organization.service.ts
```diff
@@ -39,15 +38,15 @@ export class OrganizationService {
     return this.httpClient.get<Organization>(OrganizationService.apiPrefix(code));
   }
 
-  public deleteOrganization(code: string): Observable<Response> {
+  public deleteOrganization(code: string): Observable<Object> {
```
justmesr: Isn't it better to use `object` instead of `Object` according to Typescript's "Do's and Don'ts" https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html? :)

livthomas: This is the default return type of several `HttpClient` methods. However, I used it just temporarily in order to fix type errors. We can discuss what is the best return type for our service methods.

PR #90, opened 2017-08-22
Allow document navigation and optimize attribute suggestions #LMR-310

justmesr:
- Fix layout performance issues and bugs (by replacing used library)
 - Optimize and stylize attribute suggestions
 - Allow keyboard navigation across documents
 - Implement `edit` and `select` modes for documents
 - Extract utility classes
 - Fix spelling mistakes


justmesr:
No edit rights (key navigation disabled, no edti anddelete buttons):
selection_004

Suggestion styles:
selection_005

Layout behavior:
peek 2017-08-23 09-15

justmesr:
Uninitialized documents:
peek 2017-08-26 12-12

src/main/scripts/src/app/app.component.ts
```diff
@@ -33,8 +33,8 @@ export class AppComponent implements OnInit {
 
   public notificationsOptions = {
     position: ['top'],
-    timeOut: 5000,
-    showProgressBar: false,
+    timeOut: 1000,
+    showProgressBar: true,
```
livthomas: What is the progress bar good for?

src/main/scripts/src/app/shared/perspectives/documents/post-it/attribute/attribute-property-selection.ts
```diff
@@ -18,12 +18,12 @@
...
 
-export interface AttributePropertyInput {
+export interface AttributePropertySelection {
 
-  id: string;
-  propertyName: string;
-  inputTableX: number;
-  inputTableY: number;
-  element: HTMLInputElement;
+  direction: string;
```
livthomas: What about using `enum` instead? You can even make use of [TypeScript 2.4 latest features](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html).

src/main/scripts/src/app/shared/perspectives/documents/post-it/document/post-it-document.component.html
```diff
@@ -1,33 +1,94 @@
...
   <span class="secondary-edit">
     <!--the comments remove added space between icons-->
-    <a class="fa fa-fw fa-sign-in" aria-hidden="true" title="Incoming Links"></a><!--
-    --><a class="fa fa-fw fa-sign-out" aria-hidden="true" title="Outgoing Links"></a><!--
-    --><a class="fa fa-fw fa-history" aria-hidden="true" title="History"></a><!--
-    --><a class="fa fa-fw fa-users" aria-hidden="true" title="User Rights"></a>
+    <a class="fa fa-fw fa-sign-in" title="Incoming Links"></a><!--
+    --><a class="fa fa-fw fa-sign-out" title="Outgoing Links"></a><!--
+    --><a class="fa fa-fw fa-history" title="History"></a><!--
+    --><a class="fa fa-fw fa-users" title="User Rights"></a>
```
livthomas: Why are there empty comments here?

src/main/scripts/src/app/shared/perspectives/documents/post-it/document/post-it-document.component.ts
```diff
@@ -18,19 +18,28 @@
...
 @Component({
   selector: 'post-it-document',
   templateUrl: './post-it-document.component.html',
   styleUrls: ['./post-it-document.component.scss']
 })
-export class PostItDocumentComponent {
+export class PostItDocumentComponent implements OnInit {
+
+  @HostBinding('style.border-color')
+  private borderColor: string;
```
livthomas: Do we really want to change border color? In my opinion, it does not look good. There is too much of the same color on the page. I think the icon and collection name in that color is enough.

src/main/scripts/src/app/shared/perspectives/documents/post-it/post-it-documents-perspective.component.html
```diff
@@ -1,26 +1,25 @@
...
-                      (attributePairChange)="onAttributePairChange(document, $event)">
+                      (changes)="sendUpdate(idx)">
     </post-it-document>
-
   </div>
 </div>
+<div class="fadeup"></div>
 
-<a (click)="increaseBlockHeight()" class="load-more">See More<i class="fa fa-chevron-down" aria-hidden="true"></i></a>
+<a (click)="onSeeMore(perspective)" class="see-more">See More<i class="fa fa-chevron-down"></i></a>
```
livthomas: Do this really need to be shown all the time? What about the situation when there are no documents in the collection? Or what if you all the documents are already shown?

src/main/scripts/src/app/utils/Popup.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------/
+ */
+
+export class Popup {
```
livthomas: We should not create another folders in the root `app` folder except for its submodules. All the utils classes should probably be moved to `shared/utils` folder or directly to the specific modules if they are not used anywhere else.

src/main/scripts/webpack.config.js
```diff
@@ -75,8 +75,7 @@ module.exports = {
     ),
     new webpack.ProvidePlugin({
       $: 'jquery',
-      jQuery: 'jquery',
-      BootstrapDialog: 'BootstrapDialog'
+      jQuery: 'jquery'
```
livthomas: I just wonder why we need jQuery in our code. Do not the latest EcmaScript methods provide most if not all functionality  of this library?

src/main/scripts/src/app/app.component.ts
justmesr: The main reason was that it looks good.
On the practical side however, by having an animation it brings more attention to the popup. Lets users know how long until it disappears, and lets them know that by hovering above the popup the timer stops (..so they can read the message).

src/main/scripts/src/app/shared/perspectives/documents/post-it/document/post-it-document.component.html
justmesr: `<!--the comments remove added space between icons-->`
Right above it. :D
Without the comments I would have to put them all on same line in order to not have a space character inserted between the icons.
(Not sure what's the exact reason it happens, but it looks like html interprets all `\n` inside `<span>` as a space)

src/main/scripts/src/app/shared/perspectives/documents/post-it/document/post-it-document.component.ts
justmesr: I and Maria liked it.. plus it looks way better on collections, so I added it to documents too, to keep the style same.
I'll ask in a `web-ui`, which one to keep.
selection_001

src/main/scripts/src/app/shared/perspectives/documents/post-it/post-it-documents-perspective.component.html
justmesr: Didn't think of that. But both cases are just not having anything more to show on the backend.. which is currently always the case.
Unless you have any suggestion, I can rename it to `scroll to bottom` not to confuse users right now. And implement the hiding alongside requesting more documents later.

src/main/scripts/webpack.config.js
justmesr: Many libraries that don't have Angular binding use it.. so it might come in handy.
AFAIK it's not used anywhere in our code right now. (besides the `See More`, for smooth scroll to the bottom of `<div>`, which will be reworked either way)

PR #91, opened 2017-08-27
Enable users to create, edit, and delete collections #LMR-270

justmesr:


justmesr:
New Logo:
selection_005
Old logo:
selection_006

Creation:
peek 2017-08-27 16-58

src/main/scripts/src/app/workspace/workspace-chooser/workspace-chooser.component.ts
```diff
@@ -107,59 +111,57 @@ export class WorkspaceChooserComponent implements OnInit {
...
-    }
+  public onScrollOrganizations(direction: number): void {
+    this.companiesElement.scrollToLeft(this.companiesElement['elementRef'].element.scrollLeft + squareSize * direction);
   }
 
-  public onScrollProjects(toRight?: boolean) {
+  public onScrollProjects(toRight?: boolean): void {
     if (toRight) {
-      this.projectsEl.scrollToLeft(this.projectsEl.elementRef.nativeElement.scrollLeft + squareSize);
+      this.projectsElement.scrollToLeft(this.projectsElement['elementRef'].element.scrollLeft + squareSize);
```
livthomas: I am not able to compile the application on my computer because of this change. When I change it back to `nativeElement`, it seems to work.

livthomas:
And I can see no error message when saving the collection on the server fails.

justmesr:
>"The placeholder text in the input field should be grey (as it is in the search bar)."
It is.. just the boldness makes it look darker.  (Making it thinner looked bad and changing placeholder color is kinda problematic.. if it's really an issue I could change it to not use placeholder at all and simulate placeholder behavior using `ngModel`)

>"The selected organization and project in the workspace chooser should be highlighted."
Only the active one has colored border.

PR #97, opened 2017-08-30
Refactor perspectives to use searchService

justmesr:

### DIFF
--- a/src/main/scripts/src/app/core/rest/collection.service.ts
+++ b/src/main/scripts/src/app/core/rest/collection.service.ts
@@ -19,7 +19,7 @@
  */
 
 import {Injectable} from '@angular/core';
-import {HttpClient, HttpErrorResponse, HttpParams} from '@angular/common/http';
+import {HttpClient, HttpErrorResponse, HttpParams, HttpResponse} from '@angular/common/http';
 
 import {WorkspaceService} from '../workspace.service';
 import {Collection} from '../dto/collection';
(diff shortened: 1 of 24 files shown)


src/main/scripts/src/app/core/rest/search.service.ts
```diff
@@ -1,47 +1,52 @@
...
 @Injectable()
 export class SearchService {
 
   constructor(private http: HttpClient, private workspaceService: WorkspaceService) {
   }
 
-  public suggest(text: string, type: string): Observable<Suggestion[]> {
-    return this.http.get<Suggestion[]>(`${this.searchPath()}/suggestion`,
+  public suggest(text: string, type: string): Observable<Suggestions> {
+    return this.http.get<Suggestions>(`${this.searchPath()}/suggestion`,
```
livthomas: `.../suggestions`

src/main/scripts/src/app/core/rest/collection.service.ts
```diff
@@ -38,11 +39,26 @@ export class CollectionService {
...
+    return this.httpClient.post(this.apiPrefix(), collection, {observe: 'response'})
       .catch(CollectionService.handleError);
   }
 
+  public updateCollection(collection: Collection): Observable<Collection> {
+    return this.httpClient.put(`${this.apiPrefix()}/${collection.code}`, collection)
+      .catch(CollectionService.handleError);
+  }
+
+  public removeCollection(collectionCode: String): Observable<HttpResponse<any>> {
```
livthomas: Haven't you said that we should not use object types? :) (`string`)

PR #101, opened 2017-09-01
Add http and keycloak providers

karelhala:
When we were migrating from old app file structure we forgot to include `keycloak-http.service.ts` and `keycloak.service.ts` and put them in providers field. It caused that HTTP request had no JWT and keycloak refused it's request.


karelhala:
@marvec @livthomas take a look at this and check if it fixes our problem. Thank you.

PR #103, opened 2017-09-02
The implementation of an abstract permissions component #LMR-247

martinglova:


src/main/scripts/src/app/core/rest/project.service.ts
```diff
@@ -19,41 +19,93 @@
...
   }
 
   public editProject(orgCode: string, projCode: string, project: Project): Observable<HttpResponse<object>> {
-    return this.httpClient.put(ProjectService.apiPrefix(orgCode, projCode), project, {observe: 'response'});
+    return this.httpClient.put(this.apiPrefix(orgCode, projCode), project, {observe: 'response'});
   }
 
-  private static apiPrefix(orgCode: string, projCode?: string): string {
+  public getPermissions(): Observable<Permissions> {
+    let orgCode = this.workspaceService.actualOrganizationCode;
```
justmesr: You can refactor all these duplicit `let code = this.workspaceService.actualOrganizationCode; let projCode = this.workspaceService.actualProjectCode;` to a `actualApiPrefix` method.

src/main/scripts/src/app/core/workspace.service.ts
```diff
@@ -26,6 +26,9 @@ export class WorkspaceService {
   public organizationCode: string;
   public projectCode: string;
 
+  public actualOrganizationCode: string;
+  public actualProjectCode: string;
```
justmesr: Isn't there a better name for the variables? I am not sure what's not actual about the `organizationCode` and `projectCode` codes.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -44,57 +48,182 @@ export class PermissionsTableComponent implements OnInit {
...
+    if (entities) {
+      return entities;
+    } else {
+      throw Error('unknown entity type');
+    }
+  }
+
+  private setPermissionsForResource(resourceType: string) {
+    let parent = this;
+    let getPermissionsFunctions = {
```
justmesr: Does the method `getPermissionsFunctions()` have to be inlined inside `getPermissionsFunctions`?
I see no reason it couldn't be used anywhere else.

Also you can remove the `let parent = this;` and use arrow functions instead of `function() {...}`

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -44,57 +48,182 @@ export class PermissionsTableComponent implements OnInit {
...
+      throw Error('unknown resorce type');
     }
   }
 
   private getOtherEntities(entityType: string): string[] {
     switch (entityType) {
       case 'users':
         // TODO get from usersService instead
-        return ['alicak', 'kubedo', 'jkotrady', 'kulexpipiens'];
+        let userEntities = ['alicak', 'kubedo', 'jkotrady', 'kulexpipiens'];
```
justmesr: Can you extract user and group entities to some mockup of `userService`.. it doesn't have to work for now, just return the array. So that when implementing the service you just have to add proper implementation to the service instead of `return Observable.of(['alicak', 'kubedo', 'jkotrady', 'kulexpipiens']);` and not look for TODOs across the project.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -44,57 +48,182 @@ export class PermissionsTableComponent implements OnInit {
...
+        }
+        return groupEntities;
       default:
         throw Error('unknown entity type');
     }
   }
 
-  public onAdd() {
-    // TODO call remote service
+  public onAdd(selectedName: string) {
```
justmesr: You can split the function into smaller parts (`getResource` and `updatePrmissions`).

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -44,57 +48,182 @@ export class PermissionsTableComponent implements OnInit {
...
+      'users': function(permission: Permission) {return resource.updateUserPermission(permission);} ,
+      'groups': function(permission: Permission) {return resource.updateGroupPermission(permission);}
+    };
+
+    let updatePermissions = updatePermissionsFunctions[this.entityType];
+    if (updatePermissions) {
+      let permission: Permission = {name: selectedName, roles: this.addedRoles};
+      updatePermissions(permission).subscribe();
+      this.entities.push(permission);
+      this.otherEntities.splice(this.otherEntities.indexOf(selectedName), 1);
```
justmesr: What are `otherEntities`? How are they 'other' from `entities`?

src/main/scripts/src/app/core/workspace.service.ts
martinglova: The difference is that the `organizationCode` and `projectCode` variables are only changed when both are changed together. But maybe there is a better name for them or a better way how to do it. Any suggestions?

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
martinglova: Not sure what you mean by this comment from the beginning...

And if I'm not mistaken `this` inside an inlined function - as it is in this case - is something different as what I want so I have to keep it in another variable. And when I only use ex. `this.organizationService.getPermissions` there is an error that `HttpClient` object was not created correctly.

src/main/scripts/src/app/core/workspace.service.ts
justmesr: I am thinking of some: 
  - Name them `currentlySelectedOrganizationCode` and `currentlySelectedProjectCode`
  - If the workspace is chosen organization, project pair.. you could remove them from the service and pass the `organizationCode` to the `OrganizationService` methods.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
justmesr: Well, the whole path with defining a mapper `resourceType` -> `getPermissions` function and then calling that function can be extracted to a method ~`getPermissionsFunctions()`~ `getResourcePermissions()`.

Yes, `this` inside anonymous function doesn't refer to the object. That's why you should change it to arrow functions:
`function() {return parent.organizationService.getPermissions();}` -> `() => this.organizationService.getPermissions()`

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
martinglova: The `entities` variable stores data about users who already have any role and the `otherEntities` variable stores data about users with no defined variables for a resource.

src/main/scripts/src/app/core/rest/group.service.ts
```diff
@@ -18,9 +18,14 @@
...
 
-  name: string;
-  roles: string[];
+import {Observable} from 'rxjs/Observable';
 
+@Injectable()
+export class GroupService {
+
+    public getGroups(): Observable<string[]> {
+        return Observable.of(['directors', 'customers']);
```
livthomas: I think this method should take `organizationCode` as a parameter since groups are defined for each organization separately.

src/main/scripts/src/app/core/rest/organization.service.ts
```diff
@@ -19,40 +19,81 @@
...
+  }
+
+  public removeGroupPermission(group: string): Observable<HttpResponse<any>> {
+    return this.httpClient.delete(`${this.actualApiPrefix()}/permissions/groups/${group}`, {observe: 'response'})
+      .catch(OrganizationService.handleGlobalError);
+  }
+
+  private actualApiPrefix(): string {
+    let orgCode = this.workspaceService.currentlySelectedOrganizationCode;
+    return `/${API_URL}/rest/organizations/${orgCode}`;
```
livthomas: You are mixing two different styles here. Already existing methods take `organizationCode` as a parameter while the new ones rely on `WorkspaceService`. I would suggest unifying their behavior and require `code` parameter in every method.

src/main/scripts/src/app/core/rest/user.service.ts
```diff
@@ -19,18 +19,13 @@
...
-  // TODO communicate with SecurityService through REST API
-
-  private static apiPrefix(organizationCode: string): string {
-    return `/${API_URL}/rest/roles/organizations/${organizationCode}/`;
-  }
+@Injectable()
+export class UserService {
 
+    public getUsers(): Observable<string[]> {
+        return Observable.of(['alicak', 'kubedo', 'jkotrady', 'kulexpipiens']);
```
livthomas: This method should return a list of user objects. `User` interface (similar to the [back-end one](https://github.com/Lumeer/engine/blob/devel/api/src/main/java/io/lumeer/api/model/User.java)) needs to be created in `dto` folder:
```typescript
export interface User {
  id: string;
  username: string;
  groups: string[];
}
```

src/main/scripts/src/app/core/rest/group.service.ts
livthomas: It should also return a list of group objects so we do not have to refactor it in the future.
```typescript
export interface Group {
  name: string;
}
```

src/main/scripts/src/app/core/workspace.service.ts
```diff
@@ -26,6 +26,9 @@ export class WorkspaceService {
   public organizationCode: string;
   public projectCode: string;
 
+  public currentlySelectedOrganizationCode: string;
+  public currentlySelectedProjectCode: string;
```
livthomas: This should be removed.

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.ts
```diff
@@ -44,57 +50,190 @@ export class PermissionsTableComponent implements OnInit {
...
+            }
+            this.otherEntities = userEntities;
+          }
+        );
+        break;
       case 'groups':
-        // TODO get from groupsService instead
-        return ['directors', 'customers'];
+        this.groupService.getGroups().subscribe(
+          (groupEntities: string[]) => {
```
livthomas: There is exactly the same block of code for users and groups here. It can be extract to a new method so we do not have unnecessary duplicates here.

PR #106, opened 2017-09-06
Optimize Document Post-it perspective LMR-329

justmesr:


src/main/scripts/src/app/core/dto/query.ts
```diff
@@ -20,7 +20,7 @@
 
 export interface Query {
 
-  collectionCodes?: string[];
+  collections?: string[];
```
livthomas: Why have you changed this? It follows the structure of DTO objects defined the back-end. See [JsonQuery](https://github.com/Lumeer/engine/blob/devel/api/src/main/java/io/lumeer/api/dto/JsonQuery.java). Does the search work after this change?

src/main/scripts/src/app/core/header/header.component.html
```diff
@@ -11,18 +11,21 @@
 
 <div class="menu">
   <div class="project" *ngIf="workspaceService.isWorkspaceSet()">
-    <a [routerLink]="['/workspace']">
+    <a tabindex [routerLink]="['/workspace']">
```
livthomas: Are these changes really needed? I think the TAB button works well even without them.

src/main/scripts/src/app/core/rest/collection.service.ts
```diff
@@ -113,9 +113,20 @@ export class CollectionService {
       .catch(CollectionService.handleGlobalError);
   }
 
+  private toDto(collection: Collection): Collection {
```
livthomas: Is this method really needed? It just creates the object with the same structure as the already existing one. Or where is the catch?

src/main/scripts/src/app/shared/permissions/table/permissions-table.component.scss
```diff
@@ -1,15 +1,17 @@
...
 
 th {
-  border: 1px solid #eeeeee;
+  border: 1px solid $grey-color;
   padding: 10px 20px;
 }
 
 td {
-  border: 1px solid #eeeeee;
+  border: 1px solid $grey-color;
```
livthomas: We should probably use only Boostrap classes for styling tables so these styles will not be needed anymore.

src/main/scripts/src/app/shared/perspectives/documents/post-it/document/post-it-document.component.ts
```diff
@@ -104,38 +66,16 @@ export class PostItDocumentComponent implements OnInit {
   }
 
   private setEventListener(): void {
-    // disable scrolling when navigating using keys
+    // disable scrolling when navigating using up and down keys
     this.content.nativeElement.addEventListener('keydown', (key: KeyboardEvent) => {
-      [37, 38, 39, 40].includes(key.keyCode) && key.preventDefault();
+      [38, 40].includes(key.keyCode) && key.preventDefault();
```
livthomas: Isn't it better to use keycode constants? I think they are defined somewhere. It is not obvious now which keys are used here. We should avoid comments in code and make it self-describing instead.

src/main/scripts/src/app/shared/perspectives/icon-picker/palette/icon/icon-palette.component.ts
```diff
@@ -33,7 +33,7 @@ export class IconPaletteComponent extends Palette {
   @Input('icon')
   public active: string;
 
-  public icons = Const.icons.slice(0, 100);
+  public icons = Const.icons.slice(0, 10);
```
livthomas: Just ten icons?

src/main/scripts/src/app/core/dto/query.ts
justmesr: I guess it doesn't if it did before.. just noticed that in constructor it takes `collections` (https://github.com/Lumeer/engine/blob/devel/api/src/main/java/io/lumeer/api/dto/JsonQuery.java#L73) so I changed it to that.

src/main/scripts/src/app/core/header/header.component.html
justmesr: `<a>` elements without `tabindex` or `href` in new Bootstrap change color to that of regular text on hover.. so instead of messing with Bootstrap styles I rather added it.

src/main/scripts/src/app/core/rest/collection.service.ts
justmesr: The collection it receives from backend has additional attributes (like the `lastTimeUsed`) that caused the update method to fail.

src/main/scripts/src/app/shared/perspectives/icon-picker/palette/icon/icon-palette.component.ts
justmesr: Just to fix the slow loading (And they don't work after the FA update anyway).. I am changing the icon picker a bit in a PR with the rest of Martin's requested changes.

PR #117, opened 2017-09-21
Use angular CLI so we can use i18n and such

karelhala:
As requested by @livthomas we need to have option to use i18n in application. Based on their documentation we should use angular CLI, however when web-ui was created angular CLI was not yet ready. This PR moves all files under new ng app, we can now create components, use i18n and such.


src/main/scripts/package.json
```diff
@@ -4,84 +4,110 @@
   "description": "Lumeer Web UI",
   "main": "index.js",
   "scripts": {
-    "build": "webpack -p",
+    "ng": "ng",
+    "i18n": "ng xi18n --output-path src/i18n",
     "start": "cross-env LUMEER_ENV=development webpack-dev-server",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "build": "webpack",
```
livthomas: Shouldn't we use Angular CLI commands instead? Or why do we use `webpack` and `webpack-dev-server` here?

src/main/scripts/src/assets/js/custom.js
```diff
@@ -13,6 +13,6 @@
   $("a[href='#']").click(function(e) {
     e.preventDefault();
   });
-
+console.log('oiiiiii');
```
livthomas: This was probably added just for debugging, wasn't it?

src/main/scripts/package.json
karelhala: We can't use angular cli for build/start and serve commands anymore, since we changed our webpack config (which has to be done, because of our settings of proxy and such). However most (probably all we need) of commands are available in script part of `package.json`

Here is described how it has been done and what it is good for https://github.com/angular/angular-cli/issues/6302

PR #118, opened 2017-09-24
 Fix workspace selection page #LMR-333

kubedo8:
I'll show demo at the next meeting (25 September 2017 at 20:00)


kubedo8:
Fixes:
- Used large button with icon for selecting workspace
- Max width is now 1000px (for 5 post-its)
- Added permissions checking
- Fixed paddings
- Fixed cutted shadow in selected resource

PR #119, opened 2017-09-27
Implement collection details

justmesr:
- [x] Migrate documents to Bootstrap 4
- [x] Attributes tab
- [ ] Events tab (Layed out)
- [x] Link Types tab
- [x] Collection config component


justmesr:
screenshot from 2017-09-27 12-06-13

screenshot from 2017-09-27 12-06-35

screenshot from 2017-09-27 12-06-42

screenshot from 2017-09-27 12-06-54

screenshot from 2017-09-27 12-07-28

screenshot from 2017-09-27 12-07-57

src/main/scripts/src/app/collection/collection-managed-role.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+import 'rxjs/add/operator/retry';
+
+@Injectable()
+export class CollectionManageRoleGuard implements CanActivate {
+
+  constructor(private collectionService: CollectionService) {
+  }
+
+  public canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
+    const collectionCode = state.url.split('/').reverse()[1];
```
livthomas: Why don't you get `collectionCode` parameter from `route` here?

src/main/scripts/src/app/collection/collection-routing.module.ts
```diff
@@ -23,13 +23,50 @@ import {RouterModule, Routes} from '@angular/router';
...
+      {
+        path: 'linktypes',
+        component: CollectionLinkTypesComponent
+      },
+      {
+        path: 'events',
+        component: CollectionEventsComponent
+      },
+      {
+        path: 'accessrights',
```
livthomas: I would use `permissions` in the path.

src/main/scripts/src/app/collection/config/access-rights/collection-access-rights.component.html
```diff
@@ -0,0 +1 @@
+<permissions [resourceType]="collection"></permissions>
```
livthomas: `[resourceType]="'collection'"`

src/main/scripts/src/app/collection/config/attribute-list/attribute-list.component.scss
```diff
@@ -0,0 +1,70 @@
+@import "../../../shared/_lumeer-variables";
+
+:host {
```
livthomas: Why? Angular take care of keeping the separate styles for different components.

src/main/scripts/src/app/shared/common.scss
```diff
@@ -5,3 +5,13 @@ a {
...
 }
+
+ul {
+  list-style-type: none;
+}
+
+input {
+  border: none;
+  background: transparent;
+  outline: none;
```
livthomas: Do we want these styles in all components?

src/main/scripts/src/app/collection/collection-managed-role.guard.ts
justmesr: I've printed both `route` and `state` objects to console, but can't find the parameter anywhere.. so I've used url instead.

src/main/scripts/src/app/collection/config/attribute-list/attribute-list.component.scss
justmesr: It didn't when I used browser for navigating between pages and adding the `:host` fixed the issue.

src/main/scripts/src/app/shared/common.scss
justmesr: Well it didn't break anything and most of the inputs we use are for editable names, (so they should look like regular text) not as much for forms.

livthomas: I would move these styles to the components where they are really needed. Moreover, try to define a class with these properties and apply it on elements where it is needed. It's better than overriding the default look of elements without styles.

src/main/scripts/package.json
```diff
@@ -74,7 +74,6 @@
     "bootstrap": "4.0.0-beta",
     "lodash": "^4.17.4",
     "minigrid": "^3.1.1",
-    "ng2-webstorage": "^1.5.0",
```
livthomas: Please, do not remove this dependency. We will need it.

livthomas:
Please, rebase your changes on `upstream/devel` so we will know if you break the AOT build or not (see [this PR](https://github.com/Lumeer/web-ui/pull/135)).

PR #133, opened 2017-10-15
Fix organization and project details page #LMR-334

SquattingPenguin:


src/main/scripts/src/app/workspace/organization/organization-settings.component.ts
```diff
@@ -17,10 +17,106 @@
...
+
+  constructor(private organizationService: OrganizationService,
+              private workspaceService: WorkspaceService,
+              private route: ActivatedRoute,
+              private router: Router,
+              private projectService: ProjectService) {
+  }
+
+  public ngOnInit(): void {
+    this.organization = new Organization();
```
justmesr: It's a good idea to initialize it with some values, so that when data gets loaded, there is no flashing happening as a result of elements appearing.

```
this.organization = {
  color: '#cccccc' // grey,
  icon: 'fa fa-exclamation-circle', // exclamation mark
    .
    .
    .
}
```

And to pair with it a method that should be used instead of checks for `undefined` or `null` value of `organization`:
```
public initialized(): boolean {
  return this.organization.color !== '#cccccc' && 
             this.organization.icon !== 'fa fa-exclamation-circle' &&
                        .
                        .
                        .
}
```

src/main/scripts/src/app/workspace/organization/organization-settings.component.ts
```diff
@@ -17,10 +17,106 @@
...
+    this.organizationService.editOrganization(this.organizationCode, this.organization).subscribe();
+  }
+
+  public updateOrganizationCode(): void {
+
+    this.organizationService.editOrganization(this.originalOrganizationCode, this.organization)
+      .subscribe((response: HttpResponse<Object>) => {
+        this.originalOrganizationCode = this.organization.code;
+        this.organizationCode = this.organization.code;
+        this.router.navigate([`/organization/${this.organization.code}`]);
```
justmesr:
```
.subscribe((response: HttpResponse<Object>) => {
        this.originalOrganizationCode = this.organization.code;
        this.organizationCode = this.organization.code;
        this.router.navigate([`/organization/${this.organization.code}`]);
        error => {
          return this.errorMessage = error;
        };
      });
```
|
v
```
.subscribe((response: HttpResponse<Object>) => {
          this.originalOrganizationCode = this.organization.code;
          this.organizationCode = this.organization.code;
          this.router.navigate(['/organization', this.organization.code]);
        },
        error => {
          this.errorMessage = error;
        }
      );
```

src/main/scripts/src/app/workspace/organization/organization-settings.component.ts
```diff
@@ -17,10 +17,106 @@
...
-  templateUrl: './organization-settings.component.html'
+  templateUrl: './organization-settings.component.html',
+  styleUrls: ['./organization-settings.component.scss']
 })
-export class OrganizationSettingsComponent {
+export class OrganizationSettingsComponent implements OnInit {
+
+  private organization: Organization;
+  private organizationCode: string;
+  private errorMessage: any;
```
justmesr: Useless variable.. we have `notificationService` now.

Every: `this.errorMessage = error` -> `this.notificationService.error('Error', 'Error simplified for user');`

src/main/scripts/src/app/workspace/project/project-settings.component.ts
```diff
@@ -17,11 +17,114 @@
...
+  private errorMessage: any;
+  private originalProjectCode: string;
+  private collectionsCount: number;
+  private projectDescriptionEditable: boolean = false;
+
+  @ViewChild('projectDescription')
+  public projectDescription: ElementRef;
+
+  constructor(private projectService: ProjectService,
+              private workspaceService: WorkspaceService,
```
justmesr: Unused service.

src/main/scripts/src/app/workspace/project/project-settings.component.html
```diff
@@ -1,13 +1,95 @@
...
     <a routerLink="permissions" routerLinkActive="active" class="nav-link">
       Permissions
     </a>
   </li>
 </ul>
+
 <router-outlet></router-outlet>
+
+<!-- modal -->
+<ng-template #modal>
```
justmesr: Mark with some `TODO` so we don't forget to implement confirmation of deleting projects and organizations.

src/main/scripts/src/app/workspace/project/project-settings.component.html
```diff
@@ -1,13 +1,95 @@
...
+    <div class="dropdown-divider"></div>
+    <icon-picker class="dropdown-item" [(icon)]="project.icon"
+                 (selectedChange)="updateProject()"></icon-picker>
+  </div>
+
+  <input class="color-class name" [(ngModel)]="project.name" (blur)="updateProject()"
+         [style.color]="project.color" title="Edit name of the project">
+
+  <div class="float-right document-count color-class ml-auto">
+    <a class="count" title="Number of collections in this project" [style.color]="project.color">
```
justmesr: These should redirect since they are `<a>`

`<a class="count" title="Number of collections in this project" [style.color]="project.color">` ->
```
<a class="count" 
    title="Number of collections in this project" 
    [routerLink]="[workspacePath(), 'collections']" 
    [style.color]="project.color">
```

src/main/scripts/src/app/workspace/organization/organization-settings.component.ts
```diff
@@ -119,4 +124,8 @@ export class OrganizationSettingsComponent implements OnInit {
       this.organizationDescription.nativeElement.focus();
     }, 50);
   }
+
+  public initialized(): boolean {
+    return this.organization.code != '' && this.organization.name != '';
```
justmesr: `!=` doesn't check for type, `!==` is preffered
(`0 != ''` -> `false`; `0 !== ''` -> `true`)

`return this.organization.code != '' && this.organization.name != '';` ->
`return this.organization.code !== '' && this.organization.name !== '';`

(Also `string != ''` is often redundant and can be replaced with `string` as it checks for empty string too)

src/main/scripts/src/app/workspace/organization/organization-settings.component.scss
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  .count{
+    font-size: 20px;
+  }
+
+.description {
+
+  &:hover .edit {
+    opacity: 100;
```
justmesr: `opacity: 1;` is enough.

src/main/scripts/src/app/workspace/project/project-settings.component.ts
```diff
@@ -46,21 +45,23 @@ export class ProjectSettingsComponent implements OnInit {
...
-              private workspaceService: WorkspaceService,
               private route: ActivatedRoute,
               private router: Router,
-              private collectionService: CollectionService) {
+              private collectionService: CollectionService,
+              private notificationService: NotificationsService) {
   }
 
   public ngOnInit(): void {
+    //this.project = {
```
justmesr: Remove if not needed.

src/main/scripts/src/app/collection/collection.module.ts
```diff
@@ -27,7 +27,6 @@ import {SimpleNotificationsModule} from 'angular2-notifications/dist';
 @NgModule({
   imports: [
     SharedModule,
-    SimpleNotificationsModule,
```
livthomas: When this is removed, the application cannot be loaded after AOT compilation.

livthomas:
Please, rebase your changes on `upstream/devel` so we will know once the AOT problems are fixed here (see [this PR](https://github.com/Lumeer/web-ui/pull/135)).

src/main/scripts/src/app/collection/collection.module.ts
justmesr: It does the opposite for me.
With the module imported in `collection.module.ts` it fails to build (both regular and AOT) and without it both run with no errors.

livthomas: Try to remove `node_modules` and run `npm install` again. Does it work now?

justmesr: Same.
Although changing `from 'angular2-notifications/dist';` to `from 'angular2-notifications';` fixed the issue as well for me.

livthomas: The changes you are suggesting do not work with AOT compilation. The current solution works for me using both JIT and AOT compilation.

livthomas: If it worked for @SquattingPenguin without removing `SimpleNotificationsModule` import, it should not be removed and we can discuss your problem somewhere else since it is not related to this PR at all. (Those changes were introduced before.)

SquattingPenguin:
>How it is possible to change the icon of organization/project if it is not set? I cannot figure out how to do that now.

Icon has to be always set, a default icon is assigned to organization/project during its creation.

src/main/scripts/src/app/workspace/organization/organization-settings.component.html
```diff
@@ -1,14 +1,71 @@
...
-  <li role="presentation" class="nav-item">
-    <a routerLink="settings" routerLinkActive="active" class="nav-link">
-      Basic Settings
-    </a>
-  </li>
-  <li role="presentation" class="nav-item">
+<div class="container-fluid">
+  <div class="row">
+    <div class="col"></div>
+    <div class="whole col-sm-9 col-md-10 col-lg-11" [style.opacity]="initialized() ? '1' : '0'">
```
livthomas: It behaves very strangely now. The larger the window is, the smaller the gaps are. It should be exactly the opposite. Try to use `col-12 col-lg-11 col-xl-10` as I recommended.

Permissions table and tabs above it should also use this layout. It does not look good at all now.

src/main/scripts/src/app/workspace/organization/organization-settings.component.html
```diff
@@ -1,14 +1,71 @@
...
+               [style.color]="organization.color" i18n-title="@@header.organization.code"
+               title="Edit code of the organization">
+
+        <i class="fa fa-times pull-right text-danger icon ml-auto" aria-hidden="true"
+           i18n-title="@@header.organization.remove" title="Remove organization"
+           *ngIf="organizationCode" (click)="onDelete()"></i>
+
+      </div>
+
+      <div class="description d-inline-flex py-3">
```
justmesr: It would be better if the description spanned across the whole screen width.
`<div class="description d-inline-flex py-3">` -> `<div class="description d-flex py-3">`
`<div class="text" [attr.contenteditable]="organizationDescriptionEditable"....` -> `<div class="text flex-grow" [attr.contenteditable]="organizationDescriptionEditable"....`

PR #138, opened 2017-10-23
LMR-369 Enable creating queries with filters

kubedo8:


src/main/scripts/src/app/shared/search-box/query-item/query-item-type.ts
```diff
@@ -17,11 +17,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import {until} from 'selenium-webdriver';
```
livthomas: Why this import?

src/main/scripts/src/app/shared/search-box/query-item/condition-query-item.ts
```diff
@@ -0,0 +1,38 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {QueryItem} from './query-item';
+import {QueryItemType} from './query-item-type';
+
+export class ConditionQueryItem implements QueryItem{
```
livthomas: What is the difference between `AttributeQueryItem` and `ConditionQueryItem`? Shouldn't it be the same thing?

src/main/scripts/src/app/shared/search-box/search-box.component.ts
```diff
@@ -32,44 +35,55 @@ import 'rxjs/add/operator/map';
...
+  private shouldFocusCondition: boolean = false;
+  private selectedQueryItem: number = -1;
+  private selectedSuggestion: number = -1;
+  private textCopy: string = '';
 
   constructor(private activatedRoute: ActivatedRoute,
               private queryItemsConverter: QueryItemsConverter,
               private router: Router,
-              private searchService: SearchService,
+              private searchService: SearchMockService,
```
livthomas: I know we agreed on using mock service. But it is not possible to get any meaningful search results at the moment.

src/main/scripts/src/app/shared/search-box/search-box.component.ts
```diff
@@ -32,44 +35,55 @@ import 'rxjs/add/operator/map';
...
+
+  @ViewChild('searchBox')
+  public input: ElementRef;
 
   public queryItems: QueryItem[] = [];
-  public currentQueryItem: QueryItem = new FulltextQueryItem('');
 
   private searchTerms = new Subject<string>();
-  public suggestions: Observable<QueryItem[]>;
+  private suggestionItems: QueryItem[] = [];
```
livthomas: Why this change? It does not make sense to me. Why do you need to modify this?

src/main/scripts/src/app/shared/search-box/search-box.component.ts
```diff
@@ -32,44 +35,55 @@ import 'rxjs/add/operator/map';
...
+export class SearchBoxComponent implements OnInit, AfterViewInit {
+
+  @ViewChildren('queryItemCondition')
+  public conditions: QueryList<ElementRef>;
+
+  @ViewChild('searchBox')
+  public input: ElementRef;
 
   public queryItems: QueryItem[] = [];
-  public currentQueryItem: QueryItem = new FulltextQueryItem('');
```
livthomas: The idea behind this was that everything added to `queryItems` must be already completed query item. Otherwise, it stays in the editing mode and is just stored in `currentQueryItem`. Why have you changed this?

src/main/scripts/src/app/shared/search-box/search-box.component.ts
```diff
@@ -224,14 +377,66 @@ export class SearchBoxComponent implements OnInit {
...
 
+  private isFullTextQueryPresented(): boolean {
+    return this.queryItems.some(item => item.type === QueryItemType.Fulltext);
+  }
+
+  public isCollectionItem(queryItem: QueryItem): boolean {
+    return queryItem.type === QueryItemType.Collection;
+  }
+
+  public hex2rgba(hex: string, opacity: number): string {
```
livthomas: You do not need to deal with the opacity at the moment. I have developed a solution for this which will be added in my next pull request.

src/main/scripts/src/app/shared/search-box/query-item/condition-query-item.ts
kubedo8: `ConditionQueryItem` is for suggesting conditions..

src/main/scripts/src/app/shared/search-box/query-item/query-item-type.ts
kubedo8: I don't know, why it is here. Maybe mistake..

src/main/scripts/src/app/shared/search-box/search-box.component.ts
kubedo8: This was needed for navigation in suggestions with arrows. So we need to know suggestion count and their content.

src/main/scripts/src/app/shared/search-box/search-box.component.ts
kubedo8: So how should I do that?

src/main/scripts/src/app/shared/search-box/search-box.component.ts
kubedo8: My idea is to put everything to `quertItems` and show some kind of warning if it's not completed.

src/main/scripts/src/app/shared/search-box/search-box.component.ts
livthomas: Isn't it possible to return real results (collection suggestions) together with mocked ones (attribute suggestions) in `SearchMockService`? If it is too much work then let it be but our application is based mostly on the navigation using the search results and it will not be possible after your changes.

livthomas: I have just realized that attribute suggestions are in fact implemented on the backend so you can make use of it.

kubedo8: OK. I will do that.

PR #142, opened 2017-10-31
Extract collection constants to separate file

justmesr:
Fix import warning

### DIFF
--- a/src/app/collection/constants.ts
+++ b/src/app/collection/constants.ts
@@ -21,4 +21,4 @@ export const COLLECTION_NO_ICON = 'fa fa-exclamation-circle';
 
 export const COLLECTION_NO_COLOR = '#cccccc';
 
-export const COLLECTION_NO_CODE = 'COLLECTION_HAS_NO_CODE';
+export const COLLECTION_NO_CODE = '';
--- a/src/app/collection/constants.ts
+++ b/src/app/collection/constants.ts
@@ -17,19 +17,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import {Attribute} from './attribute';
-import {Permissions} from './permissions';
+export const COLLECTION_NO_ICON = 'fa fa-exclamation-circle';
 
-export interface Collection {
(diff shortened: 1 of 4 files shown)


src/app/collection/constants.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export const COLLECTION_NO_ICON = 'fa fa-exclamation-circle';
+
+export const COLLECTION_NO_COLOR = '#cccccc';
+
+export const COLLECTION_NO_CODE = 'COLLECTION_HAS_NO_CODE';
```
livthomas: I just wonder why we need this constant. Would not it be better if the default collection code was `null` or empty string and the server generated it? (It should be capable of it now.)

justmesr: Yes, that is the default behaviour, the `COLLECTION_NO_CODE` constant is there only to represent the default uninitialized collection in a map (and I wanted the map to be of type `string -> string` instead of `string | null -> string`).

justmesr:
The whole point of the service is to share the same collection object (using the same reference) across components that can't communicate.

Not having default collection would mean they would have to synchronize in order to fetch the collection just once and both select that one.

src/app/collection/constants.ts
livthomas: Then change its value to an empty string. It would better represent an uninitialized collection.

livthomas:
Sure, I see. But would not it be easier if the selected collection was just bind to all those child components using `@Input()` field? The changes inside that objects would be propagated to all other components and the change of the whole object could be managed by events.

I do not say your solution is wrong nor do I say that it needs to be rewritten. I just think the service is not really needed. I had to deal with the same synchronization problem in the table perspective and I managed it using the input properties.

justmesr:
It's because the tabs are not children switched by an `*ngIf` they are switched using router.. so the only way I can send them data is using params or services (https://stackoverflow.com/a/34364618/5521670).

PR #143, opened 2017-11-02
Upgrade to Angular 5.0 #LMR-437

livthomas:
We still need to wait for [angular2-notifications](https://github.com/flauc/angular2-notifications) to be fixed (or use some alternative as we planned). The AOT build is failing now because of this library.

It would also be nice to use the final version of [ngx-bootstrap](https://github.com/valor-software/ngx-bootstrap) in the future or better remove this dependency.

And there are many warnings probably because for some strange reason WebPack does not know `sourcemap` property that should be used in order to properly configure [AngularCompilerPlugin](https://github.com/angular/ngtools-webpack-builds).


livthomas:
I have fixed the AOT build but that are still some warnings that should be fixed before this pull request is merged.

PR #149, opened 2017-11-08
Update webpack and ngTools

karelhala:
### Fixes build process
* Introduces tslint checks back
  * And fixes old errors
* Uses newer version of https://github.com/angular/ngtools-webpack-builds
* Updates `package.json` to use newer versions of most of our packages
* Loading of `font-awesome` done trough index rather than trough build process.
* Adds custom changes to some plugins which we are using so further update to ngtools are easier

### Visual changes
* None

### Additional changes
* Run `npm run build` so font awesome files are correctly loaded


package.json
```diff
@@ -34,74 +34,76 @@
...
     "tslint": "^5.8.0",
-    "tslint-loader": "^3.3.0",
-    "typescript": "~2.4.0",
-    "typings": "2.1.0",
-    "url-loader": "^0.5.7",
-    "webpack-dev-server": "~2.9.1",
-    "webpack-node-externals": "^1.6.0",
-    "webpack": "~3.6.0"
+    "tslint-loader": "^3.5.3",
+    "typescript": "~2.6.1",
```
livthomas: Do we really want to update to TypeScript 2.6? Angular 5.0 officially supports TypeScript 2.4. See the warnings in the output of `npm install`.

package.json
```diff
@@ -25,95 +25,99 @@
...
     "parchment": "^1.1.3",
-    "popper.js": "1.12.3",
     "quill": "^1.3.4",
-    "quill-delta-to-html": "^0.5.1",
+    "quill-delta-to-html": "^0.5.5",
+    "angular2-notifications": "^0.9.7",
+    "bootstrap": "4.0.0-beta.2",
+    "core-js": "^2.5.3",
+    "lodash": "^4.17.4",
+    "ngx-bootstrap": "^2.0.0-rc.0",
```
livthomas: Some dependencies such as `angular2-notifications` or `ngx-bootstrap` have already been removed from this project.

package.json
```diff
@@ -25,95 +25,99 @@
...
     "tslint": "^5.8.0",
-    "tslint-loader": "^3.3.0",
-    "typescript": "~2.5.3",
-    "typings": "2.1.0",
-    "url-loader": "^0.5.7",
-    "webpack": "~3.6.0",
-    "webpack-dev-server": "~2.9.1",
-    "webpack-node-externals": "^1.6.0"
+    "tslint-loader": "^3.5.3",
+    "typescript": "2.6.2",
```
livthomas: Is it a good idea to upgrade to the latest version of TypeScript which is not officially supported by Angular 5.1? (It is planned for Angular 5.2)

src/app/view/perspectives/smartdoc/text/attribute.blot.ts
```diff
@@ -17,21 +17,22 @@
...
 
 export class AttributeBlot extends Embed {
 
-  public static blotName = 'attribute';
-  public static className = 'attribute';
-  public static tagName = 'span';
+  public blotName = 'attribute';
+  public className = 'attribute';
+  public tagName = 'span';
+  public domNode;
```
livthomas: It seems that these changes prevents this custom blot from being used in Quill editor and cause an exception at runtime. Quill is expecting these fields to be static. I have managed to fix it by referring to them using `super.field` but it is not the best solution.

Do you have any idea how we can bring typed Quill classes here so our compiler won't complain? Here is the class hierarchy of the embed blot:

- [EmbedBlot](https://github.com/quilljs/parchment/blob/master/src/blot/embed.ts)
- [LeafBlot](https://github.com/quilljs/parchment/blob/master/src/blot/abstract/leaf.ts)
- [ShadowBlot](https://github.com/quilljs/parchment/blob/master/src/blot/abstract/shadow.ts)

PR #153, opened 2017-11-13
Create document search results page #LMR-423

kubedo8:


src/app/shared/slider/slider.component.ts
```diff
@@ -0,0 +1,74 @@
...
+import {Component, ElementRef, EventEmitter, Input, OnInit, Output, ViewChild} from '@angular/core';
+import {SizeType} from './size-type';
+
+
+@Component({
+  selector: 'slider',
+  templateUrl: './slider.component.html',
+  styleUrls: ['./slider.component.scss']
+})
+export class SliderComponent implements OnInit {
```
livthomas: Just a suggestion. Maybe a better name would be `SizeSlider` instead of just `Slider` since now it sounds like general slider component which it is not.

PR #155, opened 2017-11-13
Fix notifications performance and bundle size #LMR-441

justmesr:
peek 2017-11-13 19-52

- Bundle size reduced by removing `ngx-bootstrap` dependency.
- New notifications allow for easy confirms and prompts
- Better preformance by invoking `ngOnViewChecked` caused by notifications 2-3 times less than the previous notification service implementation


src/app/view/perspectives/table/table-perspective.component.ts
```diff
@@ -24,15 +24,13 @@ import {CollectionService} from '../../../core/rest/collection.service';
 import {Collection} from '../../../core/dto/collection';
 import {Attribute} from '../../../core/dto/attribute';
 import {Query} from '../../../core/dto/query';
-import 'rxjs/add/observable/combineLatest';
```
livthomas: You must not remove this. The table cannot be loaded now.

justmesr:
peek 2017-11-14 14-13

PR #163, opened 2017-11-18
Show no search results alert #LMR-463

kubedo8:


src/app/shared/utils/query-converter.ts
```diff
@@ -34,4 +35,11 @@ export class QueryConverter {
     return queryObject;
   }
 
+  // TODO remove this method when links will be implemented in backend
+  public static removeLinksFromQuery(query: Query): Query{
+    let queryCopy = _.clone(query);
```
livthomas: I think we should get rid of `lodash` dependency. Use [Object.assign()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) instead:
```typescript
let queryCopy = Object.assign({}, query);
```

PR #178, opened 2017-12-01
Prepare application store module #LMR-502

livthomas:
I have completed the store implementation so we can use it when implementing new components as well as stabilizing the existing ones.


src/app/core/store/collections/collections.action.ts
```diff
@@ -0,0 +1,240 @@
...
+ */
+
+import {Action} from '@ngrx/store';
+import {QueryModel} from '../navigation/query.model';
+import {AttributeModel, CollectionModel} from './collection.model';
+import {PermissionModel, PermissionType} from '../permissions/permissions.model';
+
+export enum CollectionsActionType {
+
+  GET = '[Collections] Get',
```
justmesr: Wouldn't be `Symbol`s better than `string`s for this? :) 
http://2ality.com/2016/01/enumify.html

src/app/core/rest/project.service.ts
```diff
@@ -45,12 +44,16 @@ export class ProjectService extends PermissionService {
     return this.httpClient.delete(this.apiPrefix(orgCode, projCode), {observe: 'response', responseType: 'text'});
   }
 
-  public createProject(orgCode: string, project: Project): Observable<HttpResponse<any>> {
-    return this.httpClient.post(this.apiPrefix(orgCode), project, {observe: 'response', responseType: 'text'});
+  public createProject(orgCode: string, project: Project): Observable<Project> {
+    return this.httpClient.post(this.apiPrefix(orgCode), project, {observe: 'response', responseType: 'text'}).pipe(
+      map(() => project) // TODO return fresh instance from the server instead
```
justmesr: `return fresh instance from the server instead` Nice :+1:

src/app/core/store/collections/collections.action.ts
justmesr: Are we using the fact that it's a `string`? If not, this could be made `Symbol` for extra bit of safety and performance.
http://www.zsoltnagy.eu/es6-symbols-and-its-use-cases/

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -162,8 +163,8 @@ export class PostItCollectionsComponent implements OnInit, OnDestroy {
     this.collectionService.createCollection(postIt.collection)
       .pipe(finalize(() => postIt.initializing = false))
       .subscribe(
-        code => {
-          postIt.collection.code = code;
+        collection => {
+          postIt.collection.code = collection.code;
           postIt.initialized = true;
           this.getCollection(postIt);
```
justmesr: A `TODO` comment could be handy, to know that once the collection returned from `createCollection` is a valid collection from server, that no 'refreshing' to sync witch server is needed.

src/app/shared/search-box/search-box.component.ts
```diff
@@ -181,12 +180,14 @@ export class SearchBoxComponent implements OnInit, AfterViewInit {
...
-        suggestions.links.push(link);
-      }
-    }
-    return Observable.of(suggestions);
+
+    return this.linkTypeService.getLinkTypes({}).pipe(
+      map(linkTypes => {
+        const linkTypeSuggestions = linkTypes.filter(linkType => linkType.name.toLowerCase().startsWith(this.text.toLowerCase()));
+        suggestions.links.push(...linkTypeSuggestions);
+        return suggestions;
```
justmesr: Why not just return `linkTypeSuggestions`?

src/app/core/store/collections/collections.action.ts
livthomas: Yes, NgRx action types actually need to be strings (see [Action](https://github.com/ngrx/platform/blob/master/modules/store/src/models.ts#L1-L3) interface which is implemented by all our actions).

In the past, it was recommended to use string constants for actions types. However, we use new version of TypeScript which supports string enums so why not make use of them?

src/app/shared/post-it-collections/post-it-collections.component.ts
livthomas: We are going to rewrite data manipulation in those components using this NgRx store. They will not use the services directly anymore so these comments are not necessary in my opinion. What do you think?

src/app/shared/search-box/search-box.component.ts
livthomas: Because we need to return all the suggestions (including the already existing ones in that array). See the removed code above.

src/app/shared/post-it-collections/post-it-collections.component.ts
justmesr: If all the `someService.subscribe()`s will be rewritten to use NgRx, then the comments won't be needed. If not, it would be nice to have a reminder about parts of code that need to be removed.

src/app/shared/search-box/search-box.component.ts
justmesr: Ok.
The `suggestions.links = [];` confused me as clearing `suggestions`. (Which just clears `suggestions.links` and leaves the rest of `suggestions` untouched.)

src/app/collection/config/collection-config.component.html
```diff
@@ -37,8 +37,8 @@
          class="count"
          title="Number of documents in this collection"
          [style.color]="collection.color"
-         [routerLink]="[workspacePath(), 'view']"
-         [queryParams]="{ perspective: 'postit', query: documentsQuery(collection.code) }">
+         [routerLink]="[workspacePath(), 'view', 'postit']"
```
justmesr: We need this change also in `search-home.component.html`. :)

src/app/core/rest/collection.service.ts
```diff
@@ -43,13 +43,14 @@ export class CollectionService extends PermissionService {
...
 
-  public createCollection(collection: Collection): Observable<string> {
+  public createCollection(collection: Collection): Observable<Collection> {
     return this.httpClient.post(
       this.apiPrefix(), this.toDto(collection),
       {observe: 'response', responseType: 'text'}
     ).pipe(
       map(response => response.headers.get('Location').split('/').pop()),
       tap(code => this.homePageService.addLastUsedCollection(code).subscribe()),
+      map(code => ({...collection, code: code})), // TODO return fresh instance from the server instead
```
kubedo8: In some of previous commits you told me not to do that (another request to server)... why it's now okay? (reaction to comment)

src/app/core/store/collections/collection.model.ts
```diff
@@ -0,0 +1,53 @@
...
+import {PermissionsModel} from '../permissions/permissions.model';
+
+export interface AttributeModel {
+
+  id: string;
+  name: string;
+
+  constraints: string[]; // TODO use complex objects instead
+  usageCount: number;
+  intermediate?: boolean;
```
kubedo8: What is meaning of this property?

src/app/core/store/collections/collections.action.ts
```diff
@@ -0,0 +1,240 @@
...
+  GET_SUCCESS = '[Collections] Get :: Success',
+  GET_FAILURE = '[Collections] Get :: Failure',
+
+  CREATE = '[Collections] Create',
+  CREATE_SUCCESS = '[Collections] Create :: Success',
+  CREATE_FAILURE = '[Collections] Create :: Failure',
```
kubedo8: What about concatenating these constants ? For example: CREATE_SUCCESS = `${CREATE} :: Success` ?

src/app/core/store/collections/collections.effects.ts
```diff
@@ -0,0 +1,201 @@
...
+        map((dtos: Collection[]) => dtos.map(dto => CollectionConverter.fromDto(dto)))
+      );
+    }),
+    map((collections: CollectionModel[]) => new CollectionsAction.GetSuccess({collections: collections})),
+    catchError((error) => Observable.of(new CollectionsAction.GetFailure({error: error})))
+  );
+
+  @Effect()
+  public getFailure$: Observable<Action> = this.actions$.ofType<CollectionsAction.GetFailure>(CollectionsActionType.GET_FAILURE).pipe(
+    tap(action => console.error(action.payload.error)),
```
kubedo8: Do we want to write errors to console?

src/app/collection/config/collection-config.component.html
livthomas: It has already been fixed in commit 05db543960d2bf32d429e80bb50c9357365b5075.

src/app/core/store/collections/collection.model.ts
livthomas: It means that the attribute is not the leaf but represents some parent in the tree. You need this when you want to have names on the columns in the table which are split into multiple subcolumns.

src/app/core/store/collections/collections.effects.ts
livthomas: In the future, we would need some centralized solution which will send these errors to our servers. In the meantime, I think logging is the only solution, or do you have any suggestion how to handle them?

src/app/core/rest/collection.service.ts
livthomas: Do you mean this spread syntax (`{...something}`)? It depends what you want to accomplish. In this case, when you call `createCollection` method and pass your `collection` as a parameter, its `code` will not be modified. But when you for example modify some of the attributes in the returned collection, they will be modified in the collection passed as a parameter since they use the same inner objects.

When using NgRx (or Redux), you try to make all of your values immutable. If you want to modify something, you make a copy of it and change the value in a new object. If you want to modify inner object, you make copy of that one as well. It means that there will be many cases when the inner objects are actually shared between multiple states but it does not matter since you are not allowed to modify them anyway.

src/app/core/store/collections/collections.action.ts
livthomas: Do you think it will make the code more readable?

src/app/core/store/collections/collection.model.ts
kubedo8: Thank you for explanation.

src/app/core/store/collections/collections.action.ts
kubedo8: I'm not sure about that. Since it's line by line, it can stay as it is.

src/app/core/store/collections/collections.effects.ts
kubedo8: No, it's OK.

src/app/core/rest/collection.service.ts
kubedo8: Thank you for explanation, but my answer was to heading to TODO comment after the line (return fresh instance from server)... But maybe it's because NgRx, isn't it?

livthomas: Oh, I see. This is about changing the back-end in order to return the instance of the newly created collection instead of getting the collection in another request as you did.

PR #181, opened 2017-12-03
Fix issues noted on meeting demo #LMR-510

justmesr:


src/app/core/top-panel/top-panel.component.html
```diff
@@ -12,12 +12,11 @@
     <div class="d-flex justify-content-end col order-lg-3">
       <div class="panel-item lead mr-2"
            *ngIf="isWorkspaceSet()">
-        <a [routerLink]="['/workspace']"
-           title="Choose workspace">
         <span>
-          {{workspace.projectCode}}@{{workspace.organizationCode}}
+          <a [routerLink]="['/workspace']" title="Choose workspace organization">{{workspace.organizationCode}}</a>
+          &nbsp;/&nbsp;
```
livthomas: Are these non-breaking spaces really useful here? I mean there are breaking spaces before and after them. Isn't that a problem? (I assume you want to prevent splitting this workspace location into two lines when the screen is small.)

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -321,20 +332,28 @@ export class PostItCollectionsComponent implements OnInit, OnDestroy {
   }
 
   public emptyQuery(): boolean {
-    return Object.values(this.query).find(value => {
-      if (value.constructor === Array) {
-        value = value.length;
-      }
+    const queryValues = Object.values(this.query);
+    const notEmptyValue = queryValues.find(this.isNotEmpty);
+    return notEmptyValue === undefined;
```
livthomas: What exactly is being done here?

src/app/shared/search-box/search-box.component.ts
```diff
@@ -65,6 +84,8 @@ export class SearchBoxComponent implements OnInit, AfterViewInit {
   private viewSubscription: Subscription;
   private storeSubscription: Subscription;
 
+  private readonly DEFAULT_COLOR = '#faeabb';
```
livthomas: Shouldn't it be either `const` outside of the class or a static field?

src/app/collection/config/tab/attribute-list/attribute-list.component.html
```diff
@@ -70,14 +71,20 @@ <h6 class="dropdown-header">Lumeer suggests:</h6>
...
-          <span class="constraint px-2 ml-1 rounded"
-                *ngFor="let constraint of attribute.constraints; let idx = index"
-                [style.background-color]="darken(constraintColor(constraint), mouseOver.attribute === attribute && mouseOver.constraint === constraint ? 40 : 0)"
-                (mouseenter)="mouseOver = {attribute: attribute, constraint: constraint}"
-                (mouseleave)="mouseOver = {attribute: null, constraint: ''}"
-                (click)="removeConstraint(attribute, idx)"
-                title="Remove Constraint"
-                i18n="@@collection.config.tab.attributes.constraint.remove">{{ constraint }}</span>
+          <div class="constraint px-2 ml-1 rounded"
+               *ngFor="let constraint of attribute.constraints; let idx = index; trackBy:trackByIndex"
```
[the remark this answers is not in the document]
justmesr: Because the `input` inside it modifies one of the array elements (the string `constraint`). So this way it ignores those changes and only minds the array indexes.

src/app/core/rest/project.service.ts
```diff
@@ -31,31 +31,59 @@ import {PermissionService} from './permission.service';
 export class ProjectService extends PermissionService {
 
   public getProjects(orgCode: string): Observable<Project[]> {
+    if (!this.hasOrganizationApiPrefix(orgCode)) {
+      return;
```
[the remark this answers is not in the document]
justmesr: I don't think it's the best solution, but it's a working one.
Why do you think it's better to do the checking in components?

>"We are basically hiding some underlying problem here"
I will try to investigate more about it.

>"Wouldn't it be better if this method throw exception in such cases?"
Fully agree and will change the `return` to `throw`.

src/app/core/store/views/views.effects.ts
```diff
@@ -33,6 +33,7 @@ import {ViewConverter} from './view.converter';
 import {ViewModel} from './view.model';
 import {ViewsAction, ViewsActionType} from './views.action';
 import {selectViewsDictionary} from './views.state';
+import {isNullOrUndefined} from "util";
```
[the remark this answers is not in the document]
justmesr: I've had it changed in File -> Settings -> Editor -> Code Style -> Typescript -> Punctuation, which then defaults to `''` intead of `""` and adds `;` at the end of statements.

src/app/core/rest/project.service.ts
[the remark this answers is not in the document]
livthomas: Because it does not make sense to call this method if the workspace is not set and the caller (component) is responsible for making sure it is not called in that state. So in my opinion, it is a problem in a component, don't you think?

src/app/shared/post-it-collections/post-it-collections.component.ts
justmesr: Figuring out if query items are empty so that the component allows creating new collections if there is nothing searched.

src/app/workspace/workspace-chooser/workspace-chooser.component.ts
```diff
@@ -62,6 +65,10 @@ export class WorkspaceChooserComponent implements OnInit {
   public activeOrgIx: number;
   public activeProjIx: number;
 
+  private organizationsState: OrganizationsState;
+
+  private projectsState: ProjectsState;
```
livthomas: Don't do this! These states are not intended to be used directly in the components. You need to use the right selector for the data you want to get from there.

src/app/workspace/workspace-chooser/workspace-chooser.component.ts
```diff
@@ -72,37 +79,94 @@ export class WorkspaceChooserComponent implements OnInit {
...
+  private getPreviouslySelectedOrganizationIndex(): number {
+    const selectedOrganizationCode = this.organizationsState.selectedOrganizationCode;
+    return this.organizations.findIndex(organization => organization.code === selectedOrganizationCode);
+  }
+
+  private selectPreviouslySelectedProject(index: number) {
+    const selectedOrganization = this.organizations[index];
+    this.getOrganizationProjects(selectedOrganization).subscribe(
+      projects => {
+        selectedOrganization.projects = projects;
```
livthomas: This is against NgRx principles of having all objects in the store immutable. If you want to change anything, you need to dispatch the right action. The object in the store can only be changed in the reducer function!

There has already been created an action to select a given project.

src/app/workspace/workspace-chooser/workspace-chooser.component.ts
```diff
@@ -72,37 +79,94 @@ export class WorkspaceChooserComponent implements OnInit {
...
+      this.workspace = workspace;
+      this.organizationsState = organizationsState;
+      this.projectsState = projectsState;
+
+      this.getOrganizations();
+    });
+  }
+
+  private getOrganizations(): void {
+    this.organizationService.getOrganizations().subscribe(
```
livthomas: If you want to rewrite this component using NgRx, you must not call REST services directly. The way of accomplishing the same thing in NgRx is to dispatch an action getting organizations (and projects) from the server and then subscribe to the content of the store in order to show those objects in the UI.

src/app/core/rest/project.service.ts
```diff
@@ -26,13 +26,14 @@ import {Project} from '../dto';
...
 import {PermissionService} from './permission.service';
+import {LumeerError} from '../error/lumeer.error';
 
 @Injectable()
 export class ProjectService extends PermissionService {
 
   public getProjects(orgCode: string): Observable<Project[]> {
     if (!this.hasOrganizationApiPrefix(orgCode)) {
-      return;
+      throw new LumeerError('Organization not set');
```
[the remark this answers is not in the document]
justmesr: >"So why do we need to deal with the workspace here?"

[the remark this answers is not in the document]
livthomas: Oh, I see. Sorry, `hasOrganizationApiPrefix` method confused me as I thought it is checking the organization name against the one from the workspace.

src/app/workspace/workspace-chooser/workspace-chooser.component.ts
livthomas: But maybe for now you can keep it like it is and address this in the other pull-request.

PR #182, opened 2017-12-04
Refactor documents search tab using NgRx

kubedo8:
Not final solution yet, but I want to talk about some problems with @livthomas


src/app/core/store/documents/document.model.ts
```diff
@@ -21,9 +21,13 @@ export interface DocumentModel {
 
   id?: string;
   collectionCode: string;
+  collectionIcon?: string;
+  collectionColor?: string;
+  collectionName?: string;
```
livthomas: Is this really needed? Isn't there any other way how to access this collection information in the template without storing them in document model? I think there should be. But if it's too complicated and the code is a mess after such a change then we should keep the whole collection object here rather than just some of its properties.

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -45,28 +46,38 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
   @ViewChild('xlTemplate') xlTempl: TemplateRef<any>;
 
   public size: SizeType;
-  public documents: SearchDocument[];
-  private routerSubscription: Subscription;
+  public documents: DocumentModel[];
```
livthomas: What about making this an observable instead?

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -45,28 +46,38 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
...
               private collectionService: CollectionService,
               private userSettingsService: UserSettingsService) {
+    this.querySubscription = this.store.select(selectQuery)
+      .subscribe(query => {
+        this.store.dispatch(new CollectionsAction.Get({query: query}));
+        this.store.dispatch(new DocumentsAction.Get({query: query}));
+      });
+    Observable.combineLatest(this.store.select(selectAllDocuments), this.store.select(selectAllCollections))
+      .subscribe((values) => {
+        const [documents, collections] = values;
```
livthomas: Cannot we put `[documents, collections]` directly to this array function parameters?

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -45,28 +46,38 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
...
+        if(document){
+          this.documents = documents;
+          if(collections){
+            this.initCollectionsInDocuments(collections);
```
livthomas: Isn't this method breaking the store objects immutability that is one of the basic concepts of NgRx? Wouldn't it be better to map the array of documents to a new array with their copies enriched by collection info?

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -99,54 +110,40 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
...
-  public createDefaultAttributeHtml(document: SearchDocument): string {
-    const data = document.document.data;
+  public createDefaultAttributeHtml(document: DocumentModel): string {
+    const data = document.data;
     return this.valueHtml(Object.values(data)[0]);
   }
 
-  public toggleDocument(document: SearchDocument) {
+  public toggleDocument(document: DocumentModel) {
     document.opened = !document.opened;
```
livthomas: You are breaking the immutability here. All changes of the store objects need to be done in the reducer only! So rather than doing this, you need to define an action which will mark the given document as opened.

However, I'm not sure if it is better to mix this in the documents store through `opened` field as you suggested or to create a new search store where there will be an array of expanded document IDs. The latter approach will have one advantage, it will allow to manage the expanded documents at one place and decide when you want to clear them. For example I think the expected behavior would be to keep them explanded when you change the search tabs but collapse them when you make a new search. What do you think?

src/app/view/perspectives/search/documents/search-documents.component.ts
livthomas: Just to make it clear. In this case it would not make sense to keep the copy of collection information in the documents store since it might become stale. That's why I'm suggesting creating enriched copies of document objects just for UI rendering so the collection information will be refreshed every time there is some change.

src/app/view/perspectives/search/documents/search-documents.component.html
```diff
@@ -1,11 +1,11 @@
-<div class="mt-2" *ngIf="documents">
-  <div *ngIf="documents.length == 0" class="alert alert-warning" i18n="@@search.result.empty">
+<div class="mt-2">
+  <div *ngIf="!((documents$ | async)?.length)" class="alert alert-warning" i18n="@@search.result.empty">
```
justmesr: You can do once `documents$ | async; let documents` and use documents from then on.

src/app/core/store/search/search.action.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Action} from '@ngrx/store';
+
+export enum SearchActionType {
+
+  GET_SELECTED_DOCUMENTS = '[Search Documents] Get selected',
```
livthomas: You don't need an action for getting selected documents since you don't have to load it from the server and so you can directly read it from the store.

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -46,33 +46,27 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
   @ViewChild('xlTemplate') xlTempl: TemplateRef<any>;
 
   public size: SizeType;
-  public documents: DocumentModel[];
+  public documents$: Observable<DocumentModel[]>;
   private querySubscription: Subscription;
 
   constructor(private searchService: SearchService,
               private store: Store<AppState>,
+              private searchStore: Store<SearchState>,
```
livthomas: Why don't you inject `Store<State>`?

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -46,33 +46,27 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
...
     let userSettings = this.userSettingsService.getUserSettings();
     this.size = userSettings.searchSize ? userSettings.searchSize : SizeType.M;
+    this.querySubscription = this.store.select(selectQuery)
+      .pipe(
+        tap(query => this.store.dispatch(new DocumentsAction.Get({query: query}))),
+        tap(query => this.searchStore.dispatch(new SearchAction.ClearSelectedDocuments()))
+      ).subscribe();
+    this.documents$ = this.store.select(selectDocumentsByQuery);
+    this.searchStore.select(selectSelectedDocuments)
+      .subscribe(selected => console.log('component', selected));
```
livthomas: I assume `console.log` will go away once you make it work. :)

src/app/core/store/search/search.action.ts
kubedo8: Yes yes you are right. This action is unused. I will delete it..

src/app/core/store/collections/collections.state.ts
```diff
@@ -20,18 +20,18 @@
...
 
 export interface CollectionsState extends EntityState<CollectionModel> {
-
+  loaded: boolean;
 }
 
 export const collectionsAdapter = createEntityAdapter<CollectionModel>({selectId: collection => collection.code});
 
-export const initialCollectionsState: CollectionsState = collectionsAdapter.getInitialState();
+export const initialCollectionsState: CollectionsState = {...collectionsAdapter.getInitialState(), loaded: false};
```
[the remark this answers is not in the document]
kubedo8: Nice, thanks!

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -57,16 +61,22 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
...
-      switchMap(query => this.searchService.searchDocuments(query)),
-    ).subscribe(documents => this.initDocuments(documents));
+    this.querySubscription = this.store.select(selectQuery)
+      .pipe(
+        tap(query => this.store.dispatch(new DocumentsAction.Get({query: query}))),
+        tap(query => this.store.dispatch(new ViewsAction.ChangeConfig({config: {search: {expandedDocumentIds: []}}})))
+      ).subscribe();
+    this.documents$ = this.store.select(selectDocumentsByQuery);
+    this.searchConfigSubscription = this.store.select(selectViewSearchConfig)
+      .subscribe(config => this.expandedDocumentIds = config.expandedDocumentIds.slice(0));
```
livthomas: You don't really need to pass `0` as an argument.

src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -77,10 +87,10 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
...
   }
 
-  public getTemplate(document: SearchDocument): TemplateRef<any> {
-    if (document.opened) {
-      return this.xlTempl;
-    }
+  public getTemplate(document: DocumentModel): TemplateRef<any> {
+     if (this.isDocumentOpened(document)) {
+       return this.xlTempl;
+     }
```
livthomas: What about using auto-formatting to avoid such changes of indentation?

PR #185, opened 2017-12-09
Use array and object destructuring

livthomas:


src/app/core/store/collections/collections.effects.ts
```diff
@@ -113,10 +110,10 @@ export class CollectionsEffects {
       const attributeDto = CollectionConverter.toAttributeDto(action.payload.attribute);
 
       return this.collectionService.updateAttribute(action.payload.collectionCode, action.payload.attributeId, attributeDto).pipe(
-        map(result => [action, CollectionConverter.fromAttributeDto(attributeDto)])
+        map(result => ({action: action, attribute: CollectionConverter.fromAttributeDto(result)}))
       );
```
kubedo8: I'm not sure, but when label and value has same name, it's not necessary to put label...

livthomas: Good point!

PR #186, opened 2017-12-10
Perspective icons, Search all & documents NgRx #LMR-503 #LMR-520

kubedo8:


src/app/view/view-controls/view-controls.component.ts
```diff
@@ -79,15 +80,36 @@ export class ViewControlsComponent implements OnInit, OnDestroy {
...
+  }
+
+  public getTitleForPerspective(perspective: string): string {
+    switch (perspective) {
+      case Perspective.Table:
+        return 'Table';
+      case Perspective.PostIt:
+        return 'Post-it';
+      case Perspective.Search:
+        return 'Search';
```
livthomas: This should rather use [select ICU expression](https://angular.io/guide/i18n#nesting-plural-and-select-icu-expressions) directly in the template in order to have it translatable.

src/app/view/view-controls/view-controls.component.ts
```diff
@@ -79,15 +80,36 @@ export class ViewControlsComponent implements OnInit, OnDestroy {
...
   public perspectives(): string[] {
     return Object.values(Perspective);
   }
 
+  public getIconForPerspective(perspective: string): string {
+    switch (perspective) {
+      case Perspective.Table:
+        return 'fa-table';
+      case Perspective.PostIt:
+        return 'fa-file';
```
livthomas: This will be a better fit for smart document perspective. What about using `fa-sticky-note` instead?

src/app/view/view-controls/view-controls.component.ts
```diff
@@ -79,15 +80,36 @@ export class ViewControlsComponent implements OnInit, OnDestroy {
...
+      case Perspective.Search:
+        return 'fa-list-ul';
```
livthomas: Wouldn't `fa-search` be a better fit here?

src/app/view/view-controls/view-controls.component.html
```diff
@@ -1,19 +1,19 @@
...
+    <div class="dropdown-toggle py-1 px-2" id="dropdownMenuButton" data-toggle="dropdown"
+         aria-haspopup="true" aria-expanded="false">
+      <i class="fa fa-fw {{getIconForPerspective(view.perspective)}}"></i>
+      {{getTitleForPerspective(view.perspective)}}
+    </div>
+    <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
+      <div class="dropdown-item" *ngFor="let perspective of perspectives()"
+           (click)="onSelectPerspective(perspective)">
+        <i class="fa fa-fw {{getIconForPerspective(perspective)}}"></i>
+        {{getTitleForPerspective(perspective)}}
```
livthomas: This will probably go into `<span>` in order to have it translatable and then we should also set `ml-1` here so it will look nicer.

src/app/view/view-controls/view-controls.component.html
```diff
@@ -1,28 +1,24 @@
...
-  <div class="form-control dropdown mx-3 p-0">
-    <div class="dropdown-toggle py-1 px-2" id="dropdownMenuButton" data-toggle="dropdown"
-         aria-haspopup="true" aria-expanded="false">
+<form class="form-inline d-flex flex-nowrap">
+  <div class="form-control dropdown mr-2 p-0" title="Perspective" i18n-title="@@view.perspective">
+    <div class="dropdown-toggle py-1 px-2 d-flex flex-nowrap align-items-center"
+         id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
       <i class="fa-fw {{getIconForPerspective(view.perspective)}}"></i>
-      <span class="ml-1">{view.perspective, select, postit {Post-it} search {Search} table {Table}}</span>
+      <span class="ml-1" i18n="@@view.perspective">{view.perspective, select, postit {Post-it} search {Search} table {Table}}</span>
```
livthomas: Cannot we use non-breaking hyphen here (`&#8209;`) for Post-it name? Because at the moment, this name is split into two lines on very small screens which is not really what we want.

src/app/view/view-controls/view-controls.component.html
```diff
@@ -1,28 +1,24 @@
...
-         class="col-form-label font-weight-bold">
-    <span i18n="@@view.name">Name</span>:
-  </label>
   <input #viewName
          id="viewName"
          name="viewName"
+         title="Name" i18n-title="@@view.name"
          [value]="view.name"
-         class="form-control-lg border-0 font-italic flex-grow-2 no-outline"
+         class="form-control-lg p-2 border-0 font-italic flex-grow no-outline"
```
livthomas: What was wrong with `flex-grow-2` so you had to create a new CSS class?

src/app/view/view-controls/view-controls.component.html
```diff
@@ -1,28 +1,24 @@
...
       <i class="fa-fw {{getIconForPerspective(view.perspective)}}"></i>
-      <span class="ml-1">{view.perspective, select, postit {Post-it} search {Search} table {Table}}</span>
+      <span class="ml-1" i18n="@@view.perspective">{view.perspective, select, postit {Post-it} search {Search} table {Table}}</span>
     </div>
     <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
       <div class="dropdown-item" *ngFor="let perspective of perspectives()"
            (click)="onSelectPerspective(perspective)">
         <i class="fa-fw {{getIconForPerspective(perspective)}}"></i>
-        <span class="ml-1">{perspective, select, postit {Post-it} search {Search} table {Table}}</span>
+        <span class="ml-1" i18n="@@view.perspective">{perspective, select, postit {Post-it} search {Search} table {Table}}</span>
```
livthomas: This collides with the ID used for _Perspective_ title. What about changing it to `@@view.perspective.name` here and a few lines above?

src/app/view/view-controls/view-controls.component.html
```diff
@@ -1,28 +1,24 @@
...
     </div>
   </div>
-  <label for="viewName"
-         class="col-form-label font-weight-bold">
-    <span i18n="@@view.name">Name</span>:
-  </label>
   <input #viewName
          id="viewName"
          name="viewName"
+         title="Name" i18n-title="@@view.name"
```
livthomas: Isn't _Name_ too abstract? I think user might be confused what name it is. What about calling it _View Name_?

kubedo8: You are right. I just copied the old version.

src/app/view/view-controls/view-controls.component.html
kubedo8: Of course...

src/app/view/view-controls/view-controls.component.html
kubedo8: Of course...

src/app/view/view-controls/view-controls.component.html
kubedo8: Maybe `white-space: nowrap` is better, isn't it?

src/app/view/view-controls/view-controls.component.scss
```diff
@@ -1,5 +1,9 @@
 @import "../../shared/variables";
 
+.nowrap{
+  white-space: nowrap;
+}
```
livthomas: Nice solution! However, Bootstrap already provides [.text-nowrap](https://getbootstrap.com/docs/4.0/utilities/text/#text-wrapping-and-overflow) class so there is no need for implementing it again.

PR #188, opened 2017-12-17
Improve project and organization creation - LMR-401

kubedo8:
This is not compelte Pull request. I just wanna discuss some things with @livthomas


src/app/core/rest/organization.service.ts
```diff
@@ -46,7 +46,8 @@ export class OrganizationService extends PermissionService {
 
   public createOrganization(organization: Organization): Observable<Organization> {
     return this.httpClient.post(this.apiPrefix(), organization, {observe: 'response', responseType: 'text'}).pipe(
-      map(() => organization) // TODO return fresh instance from the server
+      map(response => response.headers.get('Location').split('/').pop()),
+      map(id => ({...organization, id: id})) // TODO return fresh instance from the server
```
livthomas: Why not changing the back-end to return the whole object instead if the change is needed anyway?

src/app/core/store/organizations/organizations.effects.ts
```diff
@@ -66,14 +68,17 @@ export class OrganizationsEffects {
 
   @Effect()
   public update$: Observable<Action> = this.actions$.ofType<OrganizationsAction.Update>(OrganizationsActionType.UPDATE).pipe(
-    switchMap(action => {
+    withLatestFrom(this.store$),
```
livthomas: Would not it be better to use selector here instead in order to get just the organizations?

src/app/core/store/organizations/organizations.effects.ts
```diff
@@ -85,9 +90,13 @@ export class OrganizationsEffects {
 
   @Effect()
   public delete$: Observable<Action> = this.actions$.ofType<OrganizationsAction.Delete>(OrganizationsActionType.DELETE).pipe(
-    switchMap(action => this.organizationService.deleteOrganization(action.payload.organizationCode).pipe(
-      map(() => action)
-    )),
+    withLatestFrom(this.store$),
```
livthomas: The same comment applies here. We should use selectors as much as possible otherwise any refactoring of the store structure will need many changes across the whole application.

I can see it in the `ProjectEffects` as well.

src/app/core/store/organizations/organizations.reducer.ts
```diff
@@ -28,11 +28,12 @@ export function organizationsReducer(state: OrganizationsState = initialOrganiza
     case OrganizationsActionType.CREATE_SUCCESS:
       return groupsAdapter.addOne(action.payload.organization, state);
     case OrganizationsActionType.UPDATE_SUCCESS:
-      return organizationsAdapter.updateOne({id: action.payload.organization.code, changes: action.payload.organization}, state);
+      console.log('reducer', action.payload);
```
livthomas: One of the console logs you have mentioned? :)

src/app/core/store/projects/projects.effects.ts
```diff
@@ -27,18 +27,20 @@ import {AppState} from '../app.state';
...
-    withLatestFrom(this.store$.select(selectProjectsOrganizationCode)),
-    skipWhile(([action, previousOrganizationCode]) => action.payload.organizationCode === previousOrganizationCode),
-    switchMap(([action]) => this.projectService.getProjects(action.payload.organizationCode).pipe(
-      map(dtos => dtos.map(dto => ProjectConverter.fromDto(dto, action.payload.organizationCode)))
-    )),
+    withLatestFrom(this.store$),
+    skipWhile(([action, state]) => !state.organizations.entities[action.payload.organizationId]),
+    switchMap(([action, state]) => {
+      const organization = state.organizations.entities[action.payload.organizationId];
+      return this.projectService.getProjects(organization.code).pipe(
```
livthomas: Nice solution!

src/app/core/store/projects/projects.effects.ts
```diff
@@ -71,15 +75,18 @@ export class ProjectsEffects {
 
   @Effect()
   public update$: Observable<Action> = this.actions$.ofType<ProjectsAction.Update>(ProjectsActionType.UPDATE).pipe(
-    switchMap(action => {
-      const organizationCode = action.payload.project.organizationCode;
+    withLatestFrom(this.store$),
+    switchMap(([action, state]) => {
+      const organization = state.organizations.entities[action.payload.project.organizationId];
+      const project = state.projects.entities[action.payload.project.id];
```
livthomas: I assume this is needed just because we want to use the old project code for the REST call. In that case I would recommend to call it either `oldProject` or `previousProjectCode` in order to be self-explaining when we modify this code in the future.

src/app/core/top-panel/top-panel.component.ts
```diff
@@ -53,15 +53,17 @@ export class TopPanelComponent implements OnInit {
   }
 
   public goToWorkspaceOrganization(): void {
-    this.store.dispatch(new OrganizationsAction.Select({organizationCode: this.workspace.organizationCode}));
-    this.store.dispatch(new ProjectsAction.Select({projectCode: null}));
+    // TODO
+    // this.store.dispatch(new OrganizationsAction.Select({organizationCode: this.workspace.organizationCode}));
+    // this.store.dispatch(new ProjectsAction.Select({projectCode: null}));
```
livthomas: How do you suggest to handle this?

src/app/core/rest/organization.service.ts
kubedo8: Yes you are right. Good catch!

src/app/core/store/organizations/organizations.effects.ts
kubedo8: Again, good catch :)))

src/app/core/store/organizations/organizations.reducer.ts
kubedo8: Haha, yes :)

src/app/core/store/projects/projects.effects.ts
kubedo8: Okay

src/app/core/top-panel/top-panel.component.ts
kubedo8: What about storing only `id`s and throw away codes and implement guard to fetch selected organization and selected project on some root component, and when some doesn't exist, redirect to `workspace`. I think we can still use codes in url, but with this way we know, that these codes really exists in DB. What do you think?

kubedo8:
@livthomas Look at the method `onFieldBlur` in `ResourceChooserComponent` (especially on lines 322-324). Of couse copy is created everytime :)

kubedo8:
Todos:
- [x] Finish my changes
- [ ] Push my commits to GitHub
- [ ] Open a pull request

src/app/core/rest/workspace.service.ts
```diff
@@ -0,0 +1,102 @@
...
+    return this.store.select(selectAllOrganizations).pipe(
+      map(organizations => organizations.find(org => org.code === code)),
+      take(1)
+    );
+  }
+
+  private getOrganizationFromApi(code: string): Observable<OrganizationModel> {
+    return this.organizationService.getOrganization(code).pipe(
+      map(organization => OrganizationConverter.fromDto(organization)),
+      tap(organization => this.store.dispatch(new OrganizationsAction.GetSuccess({organizations: [organization]}))),
```
livthomas: Why are we using services outside of the store? We must never do that! The line above can be replaced by a single action so everything is handled inside the store.

I know we want to handle errors here but there must be another way that does not violate basic NgRx principles.

src/app/core/top-panel/top-panel.component.ts
```diff
@@ -83,7 +84,7 @@ export class TopPanelComponent implements OnInit {
       if (organization && project) {
         this.store.dispatch(new OrganizationsAction.Select({organizationId: organization.id}));
         this.store.dispatch(new ProjectsAction.Select({projectId: selectProject ? project.id : null}));
-        this.router.navigate(['/workspace']);
+        this.store.dispatch(new RouterAction.Go({path: ['workspace']}))
```
livthomas: You don't have to use this action in components. It is there for navigating inside the store where you change one type of action to another. We don't need to use actions for everything.

src/app/workspace/workspace-select.guard.ts
```diff
@@ -0,0 +1,114 @@
...
+                     private workspaceService: WorkspaceService,
+                     private userSettingsService: UserSettingsService,
+                     private store: Store<AppState>) {
+  }
+
+  public canActivate(next: ActivatedRouteSnapshot,
+                     state: RouterStateSnapshot): Observable<boolean> {
+
+    return this.isSomethingSelected().pipe(
+      switchMap(selected => this.checkOrFetch(selected))
```
livthomas:
```javascript
switchMap(selected => selected || this.fetchDefault())
```

Wouldn't this be simpler?

src/app/core/store/app-state-action.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Action} from '@ngrx/store';
+
+export enum AppStateActionType {
+
+  RESET_WITHOUT_WORKSPACE = 'Reset without workspace'
```
livthomas: What about naming it `CLEAR_WORKSPACE`? I think it would be more descriptive.

src/app/core/store/app-store.module.ts
```diff
@@ -76,9 +86,28 @@ const effects = [
   ViewsEffects
 ];
 
+function appStateReducer(reducer) {
```
livthomas: Cannot we move this to `app.reducer.ts` file?

src/app/core/store/app-store.module.ts
```diff
@@ -76,9 +86,28 @@ const effects = [
...
+      };
+    }
+    return reducer(state, action);
+  }
+}
+
 @NgModule({
   imports: [
-    StoreModule.forRoot(reducers, {initialState: initialAppState}),
+    StoreModule.forRoot(reducers, {initialState: initialAppState, metaReducers: [appStateReducer]}),
```
livthomas: Why is it meta reducer?

src/app/core/store/app-state-action.ts
```diff
@@ -0,0 +1,36 @@
...
+
+}
+
+export namespace AppStateAction {
```
livthomas: In order to be concise across the application I would suggest renaming this to `AppAction` as well as renaming the enum above and the file name to `app.action.ts`.

src/app/core/rest/workspace.service.ts
kubedo8: Look at official ngrx example-app https://github.com/ngrx/example-app/blob/master/src/app/guards/book-exists.ts

src/app/core/store/app-store.module.ts
kubedo8: For some strange mysterious reason, if I move it to separate file, it doesn't work.

src/app/workspace/workspace-select.guard.ts
kubedo8: It must return observable :)

livthomas: Hmm, I see.

src/app/core/rest/organization.service.ts
```diff
@@ -56,7 +55,7 @@ export class OrganizationService extends PermissionService {
       this.apiPrefix(code), organization,
       {observe: 'response', responseType: 'text'}
     ).pipe(
-      map(() => organization) // TODO return fresh instance from the server
+      switchMap(() => this.getOrganization(organization.code))
```
livthomas: Why do we make two REST calls here?

src/app/core/rest/project.service.ts
```diff
@@ -77,7 +76,7 @@ export class ProjectService extends PermissionService {
       observe: 'response',
       responseType: 'text'
     }).pipe(
-      map(() => project) // TODO return fresh instance from the server instead
+      switchMap(id => this.getProject(orgCode, project.code))
```
livthomas: And here as well.

livthomas:
I think there are two small things that can still be improved:
1. Buttons should only be shown on hover regardless of whether the resource is selected or not.
2. When a resource is not selected, you should select it by clicking on anything inside its post-it and not start editing its name/code without selection.
But we probably do not need to do them now.

PR #189, opened 2017-12-18
Enable turning off error notifications #LMR-515

livthomas:


src/app/core/notifications/notification.service.ts
```diff
@@ -20,35 +20,56 @@
...
 
   public simple(message: string): void {
+    if (this.areNotificationsDisabled()) {
+      return;
+    }
     this.zone.runOutsideAngular(() => this.notifications.simple(message, 'Hey'));
   }
 
   public success(message: string): void {
+    console.log('success');
```
justmesr: Is this intentional?

livthomas: Oh, no. Thanks.

PR #193, opened 2017-12-25
Enable users to change order of post-its #LMR-526

justmesr:


src/app/view/perspectives/post-it/post-it-perspective.component.ts
```diff
@@ -110,22 +115,18 @@ export class PostItPerspectiveComponent implements OnInit, OnDestroy {
...
+      this.layout = new PostItLayout('.post-it-document-layout', new PostItLayoutConfig(), this.zone);
+  }
 
-    this.layout = new PostItLayout({
-      container: '.post-it-document-layout',
-      item: '.layout-item',
-      gutter: 10
-    });
+  private setCurrentCollection() {
+    this.currentCollection = (this.query && this.query.collectionCodes) ? this.query.collectionCodes[0] : null;
```
livthomas: Would not this reintroduce the issue that was there before? What will happen if there are documents from multiple collections shown in this perspective?

src/polyfills.ts
```diff
@@ -60,7 +60,9 @@ import 'core-js/es7/reflect';
  * Needed for: All but Chrome, Firefox, Edge, IE11 and Safari 10
  */
 // import 'intl';  // Run `npm install --save intl`.
-window['Minigrid'] = require('../node_modules/minigrid/dist/minigrid.min.js');
+
+window['Hammer'] = require('../node_modules/hammerjs/hammer.min.js');
+window['Muuri'] = require('../node_modules/muuri/muuri.min.js');
```
livthomas: Isn't `vendor.ts` better place for such things?

src/app/view/perspectives/post-it/post-it-perspective.component.ts
justmesr: If there are multiple the add button will not show at all.

This in fact introduced a new issue. After loading the last page of documents. The newly loaded pages are empty.. so their array of CollectionCodes was empty. Making `currentCollection === null`

justmesr:
@livthomas Somehow playing a bit with default intervals and durations fixed the performance issue. :)

src/app/shared/utils/layout/post-it-layout.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.layoutEndCallback = onMostRecentLayoutEnd;
+  }
+
+  private containerExists(): boolean {
+    return !!(document.querySelector(this.containerClassName));
+  }
+
+  public destroy(): void {
+    this.layout.off('layoutEnd', this.layoutEndCallback());
+    this.layout.destroy();
```
livthomas: `TypeError: Cannot read property 'off' of undefined`

src/vendor.ts
```diff
@@ -1 +1,2 @@
+import '../node_modules/hammerjs/hammer.min.js';
```
livthomas: `import 'hammerjs';` should suffice. Cannot be `Muuri` imported the same way?

src/app/view/perspectives/post-it/post-it-perspective.component.ts
```diff
@@ -273,6 +276,10 @@ export class PostItPerspectiveComponent implements OnInit, OnDestroy {
   }
 
   private addDocumentsToLayoutAndGetTheirCollections(documents: Document[]): void {
+    if (documents.length === 0) {
+      this.allLoaded = true;
```
livthomas: This will introduce some bugs. You are setting `allLoaded` only when an empty list of documents is received but checking it always in the condition a few lines above.

src/vendor.ts
justmesr: Tried. Typescript didn't like it.

src/app/view/perspectives/post-it/post-it-perspective.component.ts
justmesr: What bugs? Once the server returns an empty page, all pages are already loaded so the client stops requesting more.

livthomas: I see. But we should start thinking in terms of reactive components since we are gradually rewriting components using NgRx. If we ignore it until refactoring then the complete rewrite of a component will be needed.

PR #199, opened 2017-12-30
Create all search results page #LMR-424

justmesr:


src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -136,24 +127,24 @@ export class PostItCollectionsComponent implements OnInit, OnDestroy {
     });
   }
 
-  public toggleFavorite(collection: Collection) {
-    this.collectionService.toggleCollectionFavorite(collection)
+  public toggleFavorite(collection: CollectionModel) {
+    this.collectionService.toggleCollectionFavorite(CollectionConverter.toDto(collection))
```
livthomas: We should not call services directly from components anymore. Try to update the collection object by dispatching an action.

Or is this some kind of workaround before favorites are implemented on the back-end? If so, this code should rather be moved to the store.

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -184,15 +176,15 @@ export class PostItCollectionsComponent implements OnInit, OnDestroy {
   public createPostIt(postIt: PostItCollectionModel): void {
     postIt.initializing = true;
 
-    this.collectionService.createCollection(postIt.collection)
+    this.collectionService.createCollection(CollectionConverter.toDto(postIt.collection))
```
livthomas: We must not use services in the components anymore! What about dispatching an action which creates a collection? This applies to the other parts of this component as well. Or was this just a quick fix and you plan to work on it in the next pull-request?

src/app/view/perspectives/search/all/search-all.component.ts
```diff
@@ -56,18 +64,48 @@ export class SearchAllComponent implements OnInit, OnDestroy {
...
       .pipe(
         skipWhile(query => isNullOrUndefined(query)),
-        tap(query => this.store.dispatch(new DocumentsAction.Get({query}))),
         tap(query => this.store.dispatch(new CollectionsAction.Get({query}))),
+        tap(query => this.store.dispatch(new DocumentsAction.Get({query}))),
         tap(query => this.store.dispatch(new ViewsAction.Get({query})))
       ).subscribe();
-    this.documents$ = this.store.select(selectDocumentsByQuery);
-    this.collections$ = this.store.select(selectCollectionsByQuery);
-    this.views$ = this.store.select(selectViewsByQuery);
```
livthomas: Why have you changed the way this was prepared to be used? Isn't it better to be able to get parts of the results instead of waiting for all the responses from the server? The previous implementation also make use of unsubscribing done by the template.

justmesr:
@livthomas The only sketch I had was this old one:
image

But since the collections and documents search perspectives are already implemented, I just used their layout.

src/app/shared/post-it-collections/post-it-collection-model.ts
```diff
@@ -17,11 +17,11 @@
...
 
-import {Collection} from '../../core/dto';
+import {CollectionModel} from '../../core/store/collections/collection.model';
 
 export class PostItCollectionModel {
 
-  public collection: Collection;
+  public collection: CollectionModel;
   public initializing: boolean;
   public initialized: boolean;
```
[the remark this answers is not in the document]
justmesr: We still need the `initializing` flag to make sure we won't dispatch `Update` or `Create` during the creation multiple times.

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -21,21 +21,25 @@ import {Component, ElementRef, Input, NgZone, OnDestroy, OnInit, QueryList, View
 import {Store} from '@ngrx/store';
 import {PostItLayoutConfig} from 'app/shared/utils/layout/post-it-layout-config';
 import {Observable} from 'rxjs/Observable';
-import {finalize, first} from 'rxjs/operators';
+import {finalize} from 'rxjs/operators';
 import {Subscription} from 'rxjs/Subscription';
 import {COLLECTION_NO_COLOR, COLLECTION_NO_ICON} from '../../collection/constants';
-import {Collection, Query} from '../../core/dto';
+import {Query} from '../../core/dto';
```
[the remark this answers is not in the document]
justmesr: Yes. Event the value I got returned was of type `QueryModel`. Thanks.

src/app/shared/post-it-collections/post-it-collections.component.ts
justmesr: This is more of a workaround. The task was to make the search-all perspective. The collection perspective changes are there only to react to query changes when searching.

src/app/view/perspectives/search/all/search-all.component.ts
justmesr: The previous implementation didn't allowed me only to use the subscription results only in certain parts of the template. I have `loading()`, `emptySearch()` methods and `*ngIf="collections.length > 0"` conditions for divider lines (so that there are only divider lines between shown perspectives) that need to know the result of all 3 subscriptions.

src/app/shared/post-it-collections/post-it-collection-model.ts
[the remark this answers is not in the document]
livthomas: You can make use of the existing properties in `CollectionModel`. Basically, what you want to do is to distinguish between three collection states:
1. just created in component, not sent to server
2. sent to server, no response yet
3. received instance from the server

The collection would in those states if the following conditions are met:
1. neither `id` nor `correlationId` set
2. `correlationId` set but `id` not set
3. `id` set

Would it be possible to implement it like this? But it is probably a task for another PR.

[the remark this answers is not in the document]
justmesr: Sure. It would be possible.

Just a question though.. the `correlationId` is `Id` for the frontend side? If I set it to some value.. the store won't change it to different value?

src/app/view/perspectives/search/all/search-all.component.ts
livthomas: Those cases can be handled differently by combining those observables to a single one or using them at several places. Sure, unless it is more complicated than your current solution. In my opinion, it is better to work just with observables otherwise someone changing your code later might think he works with component state which is not being changed during its lifecycle.

You do not have to change this implementation now but think about it the next time you're either writing a new component or refactoring an existing one. ;)

src/app/shared/post-it-collections/post-it-collection-model.ts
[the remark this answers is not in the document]
livthomas: `id` is set by the server when the collection is actually created in the database. `correlationId` is just a front-end identifier which is not propagated to the server. It is not created automatically. You have to create it when dispatching an action to create a new collection and it will be kept in the store with that collection (even when the fresh instance from the server is received) unless you clear the store (application refresh, change of project, etc.).

PR #206, opened 2018-01-10
Collection and post it fixes #LMR-568

justmesr:
It contains refactor documents PR (merge after that, pushed it here so it is ready for review).


src/app/view/perspectives/post-it/util/selection-helper.ts
```diff
@@ -246,7 +246,7 @@ export class SelectionHelper {
   }
 
   private lastRow(): number {
-    return Object.keys(this.selectedPostIt.document.data).length;
+    return Object.entries(this.selectedPostIt.document.data).length;
```
livthomas: Doesn't it return the same result?

src/app/view/perspectives/post-it/document/post-it-document.component.ts
```diff
@@ -44,6 +44,14 @@ import Update = DocumentsAction.Update;
 })
 export class PostItDocumentComponent implements OnInit, AfterViewInit, OnDestroy {
 
+  @HostListener('focusout')
+  public onFocusOut(): void {
+    if (this.changed) {
+      this.changed = false;
+      this.changes.emit();
```
livthomas: It seems that you have changed the way post-its are saved. However, it didn't solve the refresh problem I previously mentioned. Wasn't this change supposed to fix it? Or what was the reason behind changing this?

src/app/view/perspectives/post-it/util/selection-helper.ts
justmesr: It does. It's a debugging leftover.. that does the same thing.

src/app/view/perspectives/post-it/document/post-it-document.component.ts
justmesr: This is a pretty significant optimalization. If you chcecked in the previous version how often it sends data without needing to.

justmesr:
>"the newly added one is not there"I'll look into that.

justmesr:
@livthomas The `UpdateDocumentData` in store removed the collectionCode from documents, causing all sorts of problems. It should work now.

src/app/core/rest/document.service.ts
```diff
@@ -64,6 +64,7 @@ export class DocumentService {
     return this.httpClient.put<Document>(`${this.apiPrefix(document.collectionCode)}/${document.id}/data`, document.data)
       .pipe(
         catchError(error => this.handleGlobalError(error)),
+        tap(returnedDocument => returnedDocument.collectionCode = document.collectionCode),
```
livthomas: I don't understand this change. You basically change the document outside of the pipeline. The correct implementation of your intentions would be using the `map` operator:
```typescript
map(returnedDocument => {...returnedDocument, collectionCode: document.collectionCode}),
```
However, it's still not what we want here. Why isn't `collectionCode` returned from the server?

justmesr:
Can you provide more info about the issue? I coundn't reproduce it under any conditions (multiple attempts on Chrome and Firefox even under mid-tier device emulation). It uses the same mechanism in the code as it did before.. once it sends 1 Create request to server it sets a `updating` flag that indicates it shouldn't be sent again.

There might be a chance extremely fast double click could create one twice..

livthomas:
Please, resolve the conflicts so we can merge it.

livthomas:
It seems it has not been resolved the right way.

justmesr:
@livthomas Just a file that I forgot to delete.

PR #214, opened 2018-01-14
Display build number in top panel menu #LMR-645

livthomas:
@marvec, you just need to set `BUILD_NUMBER` environment variable when building the project.


livthomas:
screencapture-localhost-7000-ui-w-lmr-town-view-search-files-1515937663779

PR #229, opened 2018-01-20
ui/w/org/proj/view/search/records - fixes

kubedo8:


src/app/view/perspectives/search/documents/search-documents.component.ts
```diff
@@ -159,7 +160,8 @@ export class SearchDocumentsComponent implements OnInit, OnDestroy {
   }
 
   private getValues(document: DocumentModel): string[] {
-    return this.getValuesFromArray(Object.values(document.data || {}));
+    if(isNullOrUndefined(document.data)) return [];
+    return this.getValuesFromArray(Object.values(document.data));
```
livthomas: Does not it behave the same way?

PR #230, opened 2018-01-20
UI - use the same delete icon throughout the system #LMR-596

justmesr:


src/app/shared/search-box/search-box.component.html
```diff
@@ -20,7 +20,7 @@
            (blur)="onConditionBlur()"
            contenteditable="true">
       </div>
-      <i class="fas fa fa-times remove" (click)="onRemoveQueryItem(ix)" aria-hidden="true"></i>
+      <i class="fas fa-trash remove" (click)="onRemoveQueryItem(ix)" aria-hidden="true"></i>
```
livthomas: I don't think we want to change it here. This button doesn't delete any data. It just removes the query item from the search box so the previously used icon is better in my opinion.

justmesr:
@livthomas solved

PR #237, opened 2018-01-21
 ui/w/$ORG/$PROJ/view/search/all - color picker partially hidden #LMR-587

justmesr:


justmesr:
@livthomas The PR can serve as a hotfix. The further changing to noneditable post-its can be done in another PR.

PR #239, opened 2018-01-25
Use links stored on the server #LMR-673

kubedo8:
I apologize, but this PR can cause some bugs. I will test it as much as I can...

### DIFF
--- a/src/app/core/rest/link-instance.service.ts
+++ b/src/app/core/rest/link-instance.service.ts
@@ -19,13 +19,14 @@
 
 import {HttpClient} from '@angular/common/http';
 import {Injectable} from '@angular/core';
+
 import {Store} from '@ngrx/store';
 import {Observable} from 'rxjs/Observable';
-import {LocalStorage} from '../../shared/utils/local-storage';
 import {LinkInstance, Query} from '../dto';
 import {AppState} from '../store/app.state';
 import {selectWorkspace} from '../store/navigation/navigation.state';
 import {Workspace} from '../store/navigation/workspace.model';
+import {switchMap} from "rxjs/operators";
 
 @Injectable()
 export class LinkInstanceService {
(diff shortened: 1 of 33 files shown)


src/app/core/rest/link-instance.service.ts
```diff
@@ -38,58 +39,26 @@ export class LinkInstanceService {
...
-  private webStorageKey(): string {
-    return `linkInstances-${this.workspace.organizationCode}/${this.workspace.projectCode}`;
+    return this.httpClient.post<LinkInstance[]>(this.restApiPrefix() + '/search', query);
   }
 
-  private restApiPrefix(collectionCode: string): string {
+  private restApiPrefix(id?: string): string {
     const organizationCode = this.workspace.organizationCode;
     const projectCode = this.workspace.projectCode;
+    const suffix = id ? `/${id}` : '';
```
livthomas: Where is this `suffix` used? It seems like a bug to me. Update and delete methods won't work without it.

src/app/core/rest/link-instance.service.ts
```diff
@@ -38,58 +39,26 @@ export class LinkInstanceService {
...
   public deleteLinkInstance(id: string): Observable<string> {
-    const linkInstances = LocalStorage.get(this.webStorageKey()) || {};
-
-    delete linkInstances[id];
-
-    LocalStorage.set(this.webStorageKey(), linkInstances);
-
-    return Observable.of(id);
+    return this.httpClient.delete(this.restApiPrefix(id))
+      .pipe(switchMap(() => Observable.of(id)));
```
livthomas: You can just use `map` operator here and you do not need to create a new observable.

src/app/shared/search-box/query-item/query-items-converter.ts
```diff
@@ -76,10 +76,15 @@ export class QueryItemsConverter {
...
       return this.linkTypeService.getLinkTypes({linkTypeIds: query.linkTypeIds}).pipe(
-        map((linkTypes: LinkType[]) => linkTypes.map(linkType => linkType.collectionCodes)),
-        map(([codes]: string[][]) => codes.filter(code => !collectionCodes.includes(code))),
-        map(codes => codes.concat(collectionCodes)),
-        switchMap(codes => this.searchService.searchCollections({collectionCodes: codes}))
+        map((linkTypes: LinkType[]) => {
+          const set = new Set<string>();
+          linkTypes.forEach(linkType => {
+            set.add(linkType.collectionIds[0]);
+            set.add(linkType.collectionIds[1]);
```
livthomas: It is not the best idea to use `Set` here since you can even create link within the same collection.

src/app/view/perspectives/smartdoc/bottom-panel/smartdoc-bottom-panel.component.ts
```diff
@@ -66,11 +66,13 @@ export class SmartDocBottomPanelComponent implements OnInit, OnDestroy {
     if (this.part.linkTypeId) {
       this.linkTypeSubscription = Observable.combineLatest(
         this.store.select(selectLinkTypeById(this.part.linkTypeId)),
-        this.store.select(selectCollectionsDictionary)
+        this.store.select(selectAllCollections).pipe(
+          map(collections => new Map(collections.map(collection => [collection.id, collection] as [string, CollectionModel])))
```
livthomas: Wouldn't it be better if this conversion was done in the subscribe method just before the data is needed? Because otherwise it will be done every time something is changed in collections store but it doesn't have to be processed later if the link is not present in the link store as we wait for it.

src/app/core/rest/link-instance.service.ts
kubedo8: Oh yes, good catch.

src/app/shared/search-box/query-item/query-items-converter.ts
kubedo8: But we want to fetch each collection just once.

src/app/view/perspectives/smartdoc/bottom-panel/smartdoc-bottom-panel.component.ts
kubedo8: Nice idea :)

PR #243, opened 2018-01-27
ui/w/$ORG/$PROJ/view/postit - new record should already contain existing attributes #LMR-600

justmesr:


src/app/view/perspectives/post-it/add-document/add-document.component.ts
```diff
@@ -52,10 +52,20 @@ export class PostItAddDocumentComponent implements OnInit, OnDestroy {
...
-      data: {}
+      data: this.dataWithAttributeNames()
     });
   }
 
+  private dataWithAttributeNames(): { [attributeName: string]: string } {
+    let result = {};
+
+    this.selectedCollection.attributes.forEach(attribute => {
+      result[attribute.name] = '';
```
livthomas: We must store all attributes under `fullName` a.k.a. `id` key otherwise there might be some conflicts.

PR #247, opened 2018-01-28
Layout user and group managemet User tab #LMR-667

justmesr:

### DIFF
--- a/src/app/core/dto/permission.ts
+++ b/src/app/core/dto/permission.ts
@@ -19,7 +19,7 @@
 
 export interface Permission {
 
-  name: string;
+  id: string;
   roles: string[];
 
 }
--- a/src/app/core/dto/user.ts
+++ b/src/app/core/dto/user.ts
@@ -22,6 +22,6 @@ export interface User {
   id?: string;
   name: string;
   email: string;
-  blocked?: boolean;
+  groups: string[];
 
 }
--- a/src/app/core/dto/user.ts
+++ b/src/app/core/dto/user.ts
@@ -20,7 +20,8 @@
 export interface User {
 
   id?: string;
-  username: string; // email
-  groups: string[];
+  name: string;
+  email: string;
+  blocked?: boolean;
 
 }
--- a/src/app/workspace/organization/users/header/organization-users-header.component.ts
+++ b/src/app/workspace/organization/users/header/organization-users-header.component.ts
@@ -18,11 +18,7 @@
  */
 
 import {Component, Input} from '@angular/core';
-import {Store} from '@ngrx/store';
 import {OrganizationModel} from '../../../../core/store/organizations/organization.model';
-import {AppState} from '../../../../core/store/app.state';
-import {UserModel} from '../../../../core/store/users/user.model';
-import {UsersAction} from '../../../../core/store/users/users.action';
 
 @Component({
   selector: 'organization-users-header',
@@ -33,17 +29,5 @@ export class OrganizationUsersHeaderComponent {
 
   @Input()
   public organization: OrganizationModel;
-
-  constructor(private store: Store<AppState>) {
-  }
-
-  public addUser(): void {
-    const user: UserModel = {
-      email: '',
-      name: '',
-    };
-
-    this.store.dispatch(new UsersAction.Create({user: user}));
-  }
 
 }
--- a/src/app/workspace/organization/users/header/organization-users-header.component.ts
+++ b/src/app/workspace/organization/users/header/organization-users-header.component.ts
@@ -1,0 +1,49 @@
+/*
+ * Lumeer: Modern Data Definition and Processing Platform
+ *
+ * Copyright (C) since 2017 Answer Institute, s.r.o. and/or its affiliates.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Component, Input} from '@angular/core';
+import {Store} from '@ngrx/store';
+import {OrganizationModel} from '../../../../core/store/organizations/organization.model';
+import {AppState} from '../../../../core/store/app.state';
+import {UserModel} from '../../../../core/store/users/user.model';
+import {UsersAction} from '../../../../core/store/users/users.action';
+
+@Component({
+  selector: 'organization-users-header',
+  templateUrl: './organization-users-header.component.html',
+  styleUrls: ['./organization-users-header.component.scss']
+})
+export class OrganizationUsersHeaderComponent {
+
+  @Input()
+  public organization: OrganizationModel;
+
+  constructor(private store: Store<AppState>) {
+  }
+
+  public addUser(): void {
+    const user: UserModel = {
+      email: '',
+      name: '',
+    };
+
+    this.store.dispatch(new UsersAction.Create({user: user}));
+  }
+
+}
--- a/src/app/workspace/organization/users/list/organization-user-list.component.html
+++ b/src/app/workspace/organization/users/list/organization-user-list.component.html
@@ -19,17 +19,27 @@
   </thead>
   <tbody>
   <ng-container *ngFor="let user of users">
+
     <organization-user
       [organization]="organization"
       [user]="user"
       [expanded]="expanded[user.email]"
-      (expandedChange)="(expanded[user.email] = !expanded[user.email])">
+      (expandedChange)="(expanded[user.email] = !expanded[user.email])"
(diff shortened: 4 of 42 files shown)


src/app/workspace/organization/users/organization-users.component.html
```diff
@@ -1 +1,151 @@
...
+                   type="text"
+                   tabindex="1"
+                   placeholder="Username"
+                   [(ngModel)]="user.name"
+                   (blur)="updateUserName(user)">
+          </td>
+
+          <td class="px-1">
+            <input class="form-control edit-border"
+                   type="email"
```
livthomas: I am not sure you should be able to modify the email since it is managed by Keycloak.

justmesr: Ok. Will it be changeable by Keycloak?

src/app/core/dto/permission.ts
```diff
@@ -19,7 +19,7 @@
 
 export interface Permission {
 
-  name: string;
+  id: string;
```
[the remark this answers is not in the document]
justmesr: Not a needed change.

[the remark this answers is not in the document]
livthomas: I think it is a step in a good direction. Just make sure you don't break anything and that it will be changed on the back-end as well.

justmesr:
@livthomas I suggest splitting into an header (with info and creation responsibility) and user-list, where each user is another component.

src/app/workspace/organization/users/header/organization-users-header.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(private store: Store<AppState>) {
+  }
+
+  public addUser(): void {
+    const user: UserModel = {
+      email: '',
+      name: '',
+    };
+
+    this.store.dispatch(new UsersAction.Create({user: user}));
```
livthomas: I don't really think it is a good idea to send an empty user to the server.

src/app/workspace/organization/users/header/organization-users-header.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  selector: 'organization-users-header',
+  templateUrl: './organization-users-header.component.html',
+  styleUrls: ['./organization-users-header.component.scss']
+})
+export class OrganizationUsersHeaderComponent {
+
+  @Input()
+  public organization: OrganizationModel;
+
+  constructor(private store: Store<AppState>) {
```
livthomas: Remember what we have discussed a while ago? Components should be split into two categories - those that use store (smart / containers) and those who do not (dump / children). See the explanation [here](https://gist.github.com/btroncone/a6e4347326749f938510#utilizing-container-components).

In this whole PR, the only smart component should probably be `OrganizationUsersComponent` and all other components should communicate with the store through it.

So in this case, you should use `@Output` property and send a new user through it. Or you don't really need to pass any argument to that event as you create an empty user and that can be done in the component manipulating the store.

src/app/workspace/organization/users/organization-users.component.html
```diff
@@ -1 +1,18 @@
-Users tab
+<div class="container" *ngIf="(organization$ | async) as organization">
+  <organization-users-header
+    [organization]="organization">
+  </organization-users-header>
+
+  <ng-container *ngIf="(users$ | async)?.length else emptyResult">
+    <organization-user-list
```
livthomas: You don't really need `ng-container` here.

src/app/workspace/organization/users/list/organization-user-list.component.html
```diff
@@ -0,0 +1,35 @@
...
+    <td class="px-2"
+        i18n="@@organizations.user.roles.manage">
+      Manage Organization
+    </td>
+    <td colspan="3">
+    </td>
+  </tr>
+  </thead>
+  <tbody>
+  <ng-container *ngFor="let user of users">
```
livthomas: Should not this be `<tr>` element?

src/app/workspace/organization/users/user-filter/organization-user-filter.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="form-control d-flex p-0 m-2 filter">
+  <input class="border-0 flex-grow-2 px-2" placeholder="Filter users"/>
+
+  <button class="btn search-button"
+          type="button"
+          title="Filter"
+          i18n-title="@@organizations.users.filter">
+    <i class="fa fa-search" aria-hidden="true"></i>
+    <span class="d-none" i18n="@@organizations.users.filter.button">Filter</span>
+  </button>
```
livthomas: Cannot we reuse [SearchButtonComponent](https://github.com/Lumeer/web-ui/blob/devel/src/app/shared/search-box/button/search-button.component.html) here so we don't write the same component twice and have unified styles across the application?

src/app/workspace/organization/users/user/organization-user.component.html
```diff
@@ -0,0 +1,67 @@
...
+     i18n-title="@@organizations.user.expand"
+     (click)="expandedChange.emit()"></i>
+</td>
+
+<td class="px-1">
+  <input class="form-control edit-border"
+         type="text"
+         tabindex="1"
+         placeholder="Username"
+         [(ngModel)]="user.name"
```
livthomas: We must NEVER EVER directly modify the objects in the store! Otherwise the whole purpose of the store is useless and we will have an application full of strange bugs.

src/app/workspace/organization/users/header/organization-users-header.component.ts
justmesr: The point of the issue was just to layout the html and css.
Adding a new user was there just for presentational purposes.

src/app/workspace/organization/users/list/organization-user-list.component.html
justmesr: No. `organization-user` and `organization-user-groups` are `<tr>`-like elements.

src/app/workspace/organization/users/user-filter/organization-user-filter.component.html
justmesr: Only if I can modify the component.
As is.. it won't be able to filter users just by typing text into it.

livthomas: I'm talking about reusing just the button not the whole search bar. It should be easy, right?

justmesr: It is.

livthomas: I'm just wondering if we really need that button here if you want to implement filtering just by typing the text. What do you think? What is the purpose of having the button here then?

justmesr: There is none.
It was copied from search bar.
I'll make it just a nonclickable icon.

src/app/workspace/organization/users/organization-users.component.ts
```diff
@@ -17,10 +17,57 @@
...
+  public organization$: Observable<OrganizationModel>;
+  public users$: Observable<UserModel[]>;
+  public groups$: Observable<GroupModel[]>;
+
+  constructor(private store: Store<AppState>) {
+  }
+
+  public ngOnInit() {
+    this.organization$ = this.store.select(selectOrganizationFromUrl);
+    this.users$ = this.store.select(selectAllUsers).pipe(tap(this.sortUsers));
```
livthomas: Why do you use `tap` operator here? It should be used only for some side effects. It is better to use `map` here so you can directly work with the result and do not rely on the sort operation being done "in the background".

src/app/workspace/organization/users/organization-users.component.ts
```diff
@@ -17,10 +17,57 @@
...
+  public onUserUpdated(user: UserModel) {
+    this.store.dispatch(new UsersAction.Update({user: user}));
+  }
+
+  public onUserDeleted(user: UserModel) {
+    this.store.dispatch(new UsersAction.Delete({userId: user.id}));
+  }
+
+  public onPermissionsChanged(organization: OrganizationModel) {
+    this.store.dispatch(new OrganizationsAction.Update({organization: organization}));
```
livthomas: This action will ignore the permissions. We will need to create new actions for updating permissions and use one of those here.

src/app/core/dto/user.ts
```diff
@@ -20,7 +20,8 @@
 export interface User {
 
   id?: string;
-  username: string; // email
+  name: string;
+  email: string;
   groups: string[];
```
livthomas: The structure of the data returned by the back-end will be slightly different. See [engine PR 125](https://github.com/Lumeer/engine/pull/125). However, you can map that to the model structure which best fits your usage needs.

src/app/workspace/organization/users/user/organization-user.component.ts
```diff
@@ -0,0 +1,133 @@
...
+    return this.hasPermission(Role.Manage);
+  }
+
+  private hasPermission(role: Role): boolean {
+    const permissions = this.userPermissions();
+    return permissions.roles.includes(role);
+  }
+
+  private userPermissions(): PermissionModel {
+    let permissions = this.organization.permissions.users.find(userPermissions => {
```
livthomas: You will probably not get all permissions with the organization unless we change the back-end implementation.

livthomas:
I know it was in the drawings but I would probably use different icon for block button. `ban` looks ok but I don't know what would be the right counterpart.

livthomas:
I am closing this since it is being worked on by @kubedo8 in #277.

PR #248, opened 2018-02-01
Refactor search box using NgRx #LMR-512

livthomas:


src/app/shared/search-box/input/search-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  public onUpAndDownArrowKeysDown(event: KeyboardEvent) {
+    event.preventDefault();
+    const direction = event.keyCode === KeyCode.UpArrow ? -1 : 1;
+    this.moveSuggestionSelection$.next(direction);
+  }
+
+  public onBackspaceKeyDown() {
+    if (!this.text) {
+      this.removeQueryItem.emit();
```
kubedo8: We should have more sophisticated condition in the future, because when we are with cursor on the begging of the text, we should allow to remove query item.

src/app/shared/search-box/input/suggestions/search-suggestions.component.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (this.moveSelectionSubscription) {
+      this.moveSelectionSubscription.unsubscribe();
+    }
+    if (this.useSelectionSubscription) {
+      this.useSelectionSubscription.unsubscribe();
+    }
+    if (this.suggestionsSubscription) {
+      this.suggestionsSubscription.unsubscribe();
+    }
+  }
```
kubedo8: Look at documentation, you can use one subscription for all observables:
https://github.com/ReactiveX/rxjs/blob/master/doc/subscription.md

src/app/shared/search-box/input/suggestions/search-suggestions.component.ts
```diff
@@ -0,0 +1,171 @@
...
+      startWith(''),
+      debounceTime(300),
+      switchMap(text => this.retrieveSuggestions(text)),
+      withLatestFrom(this.store.select(selectAllCollections)),
+      switchMap(([suggestions, collections]) => SuggestionsConverter.convertSuggestionsToQueryItems(suggestions, collections)),
+      map(queryItems => this.filterViewQueryItems(queryItems)),
+      map(queryItems => this.addFulltextSuggestion(queryItems)),
+      map(queryItems => this.filterUsedQueryItems(queryItems)),
+      catchError(error => {
+        console.error(error);
```
kubedo8: Maybe you forgot to remove this :)

src/app/shared/search-box/query-item/query-item-icons/query-item-icons.component.html
```diff
@@ -0,0 +1,16 @@
...
+    </i>
+    <ng-container *ngSwitchCase="2">
+      <i [style.color]="queryItem.colors[0]"
+         class="{{queryItem.icons[0]}} fa-xs fa-fw align-top">
+      </i><!-- TODO remove space somehow
+     --><i [style.color]="queryItem.colors[1]"
+           class="{{queryItem.icons[1]}} fa-xs fa-fw align-bottom"></i>
+    </ng-container>
+  </ng-container>
+</ng-container>
```
kubedo8: This is very nice solution 👍

kubedo8:
What a commit! Very nice piece of work 👍 👍

src/app/shared/search-box/input/suggestions/search-suggestions.component.ts
livthomas: Interesting. But I'm not sure if we should use it. Look at [this discussion](https://github.com/ReactiveX/rxjs/issues/2769). I don't really know which way we can use it safely.

src/app/shared/search-box/input/suggestions/search-suggestions.component.ts
livthomas: Not really. It was there before and I think it can stay there so we are able to see the error at least in the console. Because we should not propagate it to the user. That's why we return an observable with an empty array here.

src/app/shared/search-box/input/search-input.component.ts
livthomas: Good point! I have changed it.

PR #253, opened 2018-02-03
ui/w/$ORG/$PROJ/view/search - multiple redraws of files #LMR-605

justmesr:
Redraws were caused by `ngFor` destroying and creating elements on store changes.
Fixed by telling it to keep objects with same `collection.id`.


src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -154,7 +155,8 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
       description: '',
       color: DEFAULT_COLOR,
       icon: DEFAULT_ICON,
-      defaultAttributeId: ''
+      defaultAttributeId: '',
+      correlationId: String(Math.floor(Math.random() * 1000000000000000) + 1)
```
kubedo8: You should use `CorrelationIdGenerator.generate()` method

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -353,6 +355,10 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
     });
   }
 
+  public trackByCollection(index: number, postIt: PostItCollectionModel): number {
+    return HashHelper.hashString(postIt.collection.id || postIt.collection.correlationId);
```
kubedo8: Is hash really needed?

justmesr: Good question. :)

src/app/shared/utils/hash-helper.ts
```diff
@@ -0,0 +1,38 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export class HashHelper {
```
livthomas: We should avoid generic names such as this. I would have never guessed correctly what this class does. I suggest naming it something like `HashCodeGenerator`.

src/app/shared/utils/hash-helper.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    if (str.length === 0) {
+      return hash;
+    }
+
+    for (let i = 0; i < str.length; i++) {
+      const character = str.charCodeAt(i);
+      hash = ((hash << 5) - hash) + character;
+      hash = hash & hash; // Convert to 32bit integer
+    }
```
livthomas: Seems like a solution copied from Stack Overflow. But there was also the [ES6 version](https://stackoverflow.com/a/34842797/3047439). ;)

justmesr: Since it happens often, I suggest keeping it this way, as this aproach should still be faster on V8.

src/app/view/perspectives/post-it/post-it-perspective.component.ts
```diff
@@ -34,6 +34,7 @@ import {DeletionHelper} from './util/deletion-helper';
 import {InfiniteScroll} from './util/infinite-scroll';
 import {NavigationHelper} from './util/navigation-helper';
 import {ATTRIBUTE_COLUMN, SelectionHelper, VALUE_COLUMN} from './util/selection-helper';
+import {HashCodeGenerator} from '../../../shared/utils/hash-helper';
```
livthomas: You just forgot to rename the file.

PR #258, opened 2018-02-03
ui/w/$ORG/$PROJ/view/postit - attribute order suddenly switched #LMR-695

justmesr:

### DIFF
--- a/src/app/view/perspectives/post-it/post-it-perspective.component.ts
+++ b/src/app/view/perspectives/post-it/post-it-perspective.component.ts
@@ -34,6 +34,7 @@ import {DeletionHelper} from './util/deletion-helper';
 import {InfiniteScroll} from './util/infinite-scroll';
 import {NavigationHelper} from './util/navigation-helper';
 import {SelectionHelper} from './util/selection-helper';
+import {isNullOrUndefined} from 'util';
 import Create = DocumentsAction.Create;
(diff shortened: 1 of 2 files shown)


src/app/view/perspectives/post-it/post-it-perspective.component.ts
```diff
@@ -272,15 +273,16 @@ export class PostItPerspectiveComponent implements OnInit, OnDestroy {
     postIt.document = documentModel;
     postIt.initialized = Boolean(documentModel.id);
 
-    if (!postIt.initialized) {
-      postIt.order = -this.postIts.length;
-    } else {
-      postIt.order = this.postIts.length;
-    }
+    const direction = this.wasCreatedPreviously(postIt) ? 1 : -1;
+    postIt.order = this.postIts.length * direction;
```
livthomas: What is this calculation all about?

justmesr: Until we have post-it orders stored. I've just assigned each post-it it's index as it's order number.

justmesr: Just a temporary, until post-it order is stored.
Direction makes sure, that post-its created in current session get moved to the front, while others are kept in their original order.

PR #259, opened 2018-02-04
ui/w/$ORG/$PROJ/view/postit - all long names must have tooltips #LMR-697

justmesr:

### DIFF
--- a/src/app/view/perspectives/post-it/document/post-it-document.component.html
+++ b/src/app/view/perspectives/post-it/document/post-it-document.component.html
@@ -1,7 +1,10 @@
 <div class="card-header d-inline-flex py-1 px-3">
   <i class="icon fa-fw {{ postItModel.document.collection.icon }} mr-1"
-     [style.color]="postItModel.document.collection.color"></i>
-  <div class="collection-name" [style.color]="postItModel.document.collection.color">
+     [style.color]="postItModel.document.collection.color"
+     title="{{ postItModel.document.collection.name }}"></i>
+  <div class="collection-name"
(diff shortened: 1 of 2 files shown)


src/app/view/perspectives/post-it/document/post-it-document.component.html
```diff
@@ -29,6 +32,7 @@
       <li *ngFor="let attributePair of attributePairs; let row = index">
         <span id="{{ perspectiveId }}{{ postItModel.index }}[0, {{ row }}]" class="attribute"
               tabindex="-1"
+              title="{{ attributePair.attribute }}"
```
livthomas: Why not simply using property binding (`[title]="attributePair.attribute"`) instead of interpolation?

src/app/view/perspectives/post-it/document/post-it-document.component.html
```diff
@@ -1,7 +1,10 @@
 <div class="card-header d-inline-flex py-1 px-3">
   <i class="icon fa-fw {{ postItModel.document.collection.icon }} mr-1"
-     [style.color]="postItModel.document.collection.color"></i>
-  <div class="collection-name" [style.color]="postItModel.document.collection.color">
+     [style.color]="postItModel.document.collection.color"
+     title="{{ postItModel.document.collection.name }}"></i>
+  <div class="collection-name"
+       [style.color]="postItModel.document.collection.color"
+       title="{{ postItModel.document.collection.name }}">
```
livthomas: Why is the color as well as title defined on both elements? Would not it work if a new parent element was created for those two and it was defined there just once?

PR #260, opened 2018-02-04
ui/workspace/ - update attribute name styles throughout the app #LMR-701

justmesr:
Default attribute styles added.
Unified document view attribute and value styles.
Smart-doc attribute styles were left unchanged.

### DIFF
--- a/src/app/view/perspectives/post-it/document/post-it-document.component.html
+++ b/src/app/view/perspectives/post-it/document/post-it-document.component.html
@@ -37,7 +37,7 @@
               (keydown.enter)="selectionHelper.setEditMode(true); selectionHelper.focus()"
               (click)="clickOnAttributePair(0, row)">
 
-          <input class="border-0"
+          <input class="border-0 {{ isDefaultAttribute(attributePair.attribute) ? 'default-attribute' : null }}"
                  [(ngModel)]="attributePair.attribute"
                  [attr.list]="suggestionListId()"
                  (keydown)="$event.stopPropagation()"
--- a/src/app/view/perspectives/post-it/document/post-it-document.component.html
+++ b/src/app/view/perspectives/post-it/document/post-it-document.component.html
@@ -37,7 +37,8 @@
               (keydown.enter)="selectionHelper.setEditMode(true); selectionHelper.focus()"
               (click)="clickOnAttributePair(0, row)">
 
-          <input class="border-0 {{ isDefaultAttribute(attributePair.attribute) ? 'default-attribute' : null }}"
+          <input class="border-0"
+                 [class.font-weight-bold]="isDefaultAttribute(attributePair.attribute)"
                  [(ngModel)]="attributePair.attribute"
                  [attr.list]="suggestionListId()"
                  (keydown)="$event.stopPropagation()"
--- a/src/app/view/perspectives/post-it/document/post-it-document.component.scss
+++ b/src/app/view/perspectives/post-it/document/post-it-document.component.scss
@@ -69,12 +69,6 @@
     font-style: italic;
   }
 
-  .default-attribute input {
-    font-style: normal;
-    font-weight: bold;
-    color: #000000;
-  }
-
   span:focus, span input:focus {
     outline: auto !important;
   }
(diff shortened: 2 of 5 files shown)


src/app/view/perspectives/post-it/document/post-it-document.component.html
```diff
@@ -37,7 +37,7 @@
               (keydown.enter)="selectionHelper.setEditMode(true); selectionHelper.focus()"
               (click)="clickOnAttributePair(0, row)">
 
-          <input class="border-0"
+          <input class="border-0 {{ isDefaultAttribute(attributePair.attribute) ? 'default-attribute' : null }}"
```
kubedo8: Isn't better to use it like `[class.default-attribute]="isDefaultAttribute(attributePair.attribute)"` ?

justmesr: I don't know. Are there any advantages?

If not, we can just choose one and stick to that. As far as I know, we use the `{{ }}` syntax more.

src/app/view/perspectives/search/documents/search-documents.component.scss
```diff
@@ -0,0 +1,18 @@
+@import "../../../../shared/variables";
+
+/deep/.search-documents-attribute {
```
livthomas: I think we should avoid using `deep` selector. Why does it need to be used here?

src/app/view/perspectives/post-it/document/post-it-document.component.html
livthomas: I think it does not really matter although I find the syntax suggested by @kubedo8 much cleaner. We use both styles across the application.

src/app/view/perspectives/post-it/document/post-it-document.component.scss
```diff
@@ -64,8 +64,15 @@
...
-    width: 80px;
+    width: 75px;
+    color: $primary;
+    font-style: italic;
+  }
+
+  .default-attribute input {
+    font-style: normal;
+    font-weight: bold;
+    color: #000000;
```
livthomas: I am not really sure if this is what @marvec meant. I think all attributes should be in italics and the default one in bold as well. But I might be wrong.

If I am right you don't really need to define some special class here and can just use:
```html
[class.font-weight-bold]="isDefaultAttribute(attributePair.attribute)"
```

src/app/view/perspectives/post-it/document/post-it-document.component.scss
```diff
@@ -76,6 +83,7 @@
     padding: 0 3px;
     width: 100%;
     pointer-events: none;
+    color: #000000;
```
livthomas: I am not sure if we really want to use pure black color anywhere in our application.

src/app/view/perspectives/search/documents/search-documents.component.scss
justmesr: html is pasted in as a string.
Angular ignores the class unless the _ngcontext matches.

livthomas: And why do we put there HTML from string? Cannot it be directly in the template?

src/app/view/perspectives/post-it/document/post-it-document.component.scss
justmesr: I've tried that. There was no way to tell which one is the default one, so I made it more visible.

src/app/view/perspectives/search/documents/search-documents.component.scss
justmesr: Probably. (I guess the recursive nature of some values was the reason @kubedo8 made it like that)
It is out of the scope of this PR, so I'll leave at that.

src/app/view/perspectives/post-it/document/post-it-document.component.scss
justmesr: We do. It's the default for text in inputs.

src/app/view/perspectives/search/documents/search-documents.component.scss
kubedo8: @justmesr yes you are right. Document may have nested attributes and values and generating html directly in the template was very slow even for document with few attributes.

PR #265, opened 2018-02-09
ui/w/$ORG/$PROJ/view/search/files - handling too long file names #LMR-589

justmesr:

### DIFF
--- a/src/app/core/store/collections/collections.effects.ts
+++ b/src/app/core/store/collections/collections.effects.ts
@@ -23,7 +23,7 @@ import {Action} from '@ngrx/store';
 import {Observable} from 'rxjs/Observable';
 import {catchError, flatMap, map, switchMap, tap} from 'rxjs/operators';
 import {Collection, Permission} from '../../dto';
-import {CollectionService, SearchService} from '../../rest';
+import {CollectionService, ImportService, SearchService} from '../../rest';
 import {LinkTypesAction} from '../link-types/link-types.action';
 import {QueryConverter} from '../navigation/query.converter';
 import {NotificationsAction} from '../notifications/notifications.action';
@@ -31,6 +31,7 @@ import {PermissionsConverter} from '../permissions/permissions.converter';
 import {PermissionType} from '../permissions/permissions.model';
 import {CollectionConverter} from './collection.converter';
 import {CollectionsAction, CollectionsActionType} from './collections.action';
+import {DEFAULT_COLOR, DEFAULT_ICON} from '../../constants';
 
 @Injectable()
 export class CollectionsEffects {
@@ -91,6 +92,27 @@ export class CollectionsEffects {
   public createFailure$: Observable<Action> = this.actions$.ofType<CollectionsAction.CreateFailure>(CollectionsActionType.CREATE_FAILURE).pipe(
     tap(action => console.error(action.payload.error)),
     map(action => new NotificationsAction.Error({message: 'Failed to create file'}))
+  );
+
+  @Effect()
+  public import$: Observable<Action> = this.actions$.ofType<CollectionsAction.Import>(CollectionsActionType.IMPORT).pipe(
+    switchMap(action => {
+      return this.importService.importFile(action.payload.format, action.payload.data, action.payload.name).pipe(
+        tap(collection => {
+          collection.icon = DEFAULT_ICON;
+          collection.color = DEFAULT_COLOR;
+        }),
+        map(collection => CollectionConverter.fromDto(collection))
+      );
+    }),
+    map(collection => new CollectionsAction.ImportSuccess({collection: collection})),
+    catchError((error) => Observable.of(new CollectionsAction.ImportFailure({error: error})))
+  );
+
+  @Effect()
+  public importFailure$: Observable<Action> = this.actions$.ofType<CollectionsAction.ImportFailure>(CollectionsActionType.IMPORT_FAILURE).pipe(
+    tap(action => console.error(action.payload.error)),
+    map(action => new NotificationsAction.Error({message: 'Failed to import file'}))
   );
 
   @Effect()
@@ -213,6 +235,7 @@ export class CollectionsEffects {
 
   constructor(private actions$: Actions,
               private collectionService: CollectionService,
+              private importService: ImportService,
               private searchService: SearchService) {
   }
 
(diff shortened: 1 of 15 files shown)


src/app/shared/post-it-collections/collection-name/post-it-collection-name.component.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    return this.COLLECTION_PLACEHOLDER_NAME;
+  }
+
+  public nameFocused(collectionName: HTMLDivElement): boolean {
+    return document.activeElement === collectionName;
+  }
+
+  public onNameChanged(newCollectionName: HTMLDivElement) {
+    this.postIt.collection.name = newCollectionName.textContent;
```
livthomas: It is not a good idea to modify internals of the object passed in as the component input. It may introduce some unwanted side effects.

What about keeping the changed name in some other private property, then compare it with the original value on blur and send it in `changed` event only if the value has really been changed?

justmesr: Can't think of a case, where this would cause issues.
But ok. I'll change it.

It will require changing post-it-collections to NgRx.. 
since this way of expecting the store to update the values doesn't work with our services.

justmesr:
@livthomas I fixed the issue. However some new issues from backend arose (as mentioned on Slack). Until they are fixed, the collections won't work under this PR.

src/app/shared/post-it-collections/collection-name/post-it-collection-name.component.ts
```diff
@@ -0,0 +1,117 @@
...
+  private readonly COLLECTION_WHITESPACE_COLLAPSED = 'nowrap';
+
+  private readonly COLLECTION_WHITESPACE_EXPANDED = 'normal';
+
+  private _postIt: PostItCollectionModel;
+
+  @Input()
+  set postIt(value: PostItCollectionModel) {
+    this._postIt = value;
+    this.extractCollectionName(value.collection);
```
livthomas: Why do you complicate this component if you do not really need the whole post-it object here? You can just have an input with the collection name here and copy its value to the input in `ngOnChanges` method.

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -82,142 +77,79 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
...
-        if (success) {
-          collection.favourite = !collection.favourite;
-        }
-      });
-  }
-
-  public hasWriteRole(collection: CollectionModel): boolean {
-    return this.hasRole(collection, Role.Write);
-  }
+    const collectionCopy = Object.assign({}, collection);
```
livthomas: +1 on not modifying the object directly. But there is an easier way of making a copy:
```typescript
const collectionCopy = {...collection};
```

justmesr:
>"If I understand it correctly, we will need to wait for the back-end fixes before merging this, right?"
Yes.

src/app/core/store/collections/collections.effects.ts
```diff
@@ -93,6 +94,27 @@ export class CollectionsEffects {
     map(action => new NotificationsAction.Error({message: 'Failed to create file'}))
   );
 
+  @Effect()
+  public import$: Observable<Action> = this.actions$.ofType<CollectionsAction.Import>(CollectionsActionType.IMPORT).pipe(
+    switchMap(action => {
+      return this.importService.importFile(action.payload.format, action.payload.data, action.payload.name).pipe(
+        tap(collection => {
+          collection.icon = DEFAULT_ICON;
+          collection.color = DEFAULT_COLOR;
```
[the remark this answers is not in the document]
justmesr: The whole modification will go away, once backend adds the values on import.

src/app/shared/post-it-collections/post-it-collection-model.ts
```diff
@@ -22,7 +22,6 @@ import {CollectionModel} from '../../core/store/collections/collection.model';
 export class PostItCollectionModel {
 
   public collection: CollectionModel;
-  public initializing: boolean;
-  public initialized: boolean;
+  public hovered: boolean;
```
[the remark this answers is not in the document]
justmesr: It means that the cursor is on top of the postIt.
Where else would you put it?

src/app/core/store/collections/collections.effects.ts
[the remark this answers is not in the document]
livthomas: I know but it doesn't mean that we should have bad code there in the meantime. I have seen incorrectly used `tap` by you many times. So I'm just saying stop using it for modifying the stream. It should not be used for that.

src/app/shared/post-it-collections/collection-name/post-it-collection-name.component.ts
justmesr: Because I use the postIt attributes in the component.

src/app/shared/post-it-collections/post-it-collections.component.ts
justmesr: Does it do a deep copy?

src/app/shared/post-it-collections/post-it-collection-model.ts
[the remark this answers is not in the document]
livthomas: Why do we need it at all? Cannot you just apply styles based on if the component is hovered or not?

src/app/shared/post-it-collections/post-it-collections.component.ts
livthomas: No, but `Object.assign()` doesn't do that neither. See [examples section on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

src/app/shared/post-it-collections/collection-name/post-it-collection-name.component.ts
livthomas: You don't really need any of its attributes in this component. Try not to over-complicate things and always think about the simplest solution (with the least amount of code).

livthomas:
You should be able to finish this since the back-end has already been changed. See https://github.com/Lumeer/engine/commit/dbce3f8cdb9ea1cdfa7f6b19c6250f199e67ba4c.

livthomas:
I am closing this since it is being further worked on by @kubedo8 in #275.

PR #269, opened 2018-02-18
 Create view search results page  #LMR-425

kubedo8:


src/app/view/perspectives/search/views/view-detail/view-detail.component.html
```diff
@@ -0,0 +1,17 @@
+<div class="card p-2 mb-1 view"
+     (click)="onClicked()">
+  <div class="d-flex flex-row">
+    <i class="fa fa-fw {{getIconForPerspective(view.perspective)}} align-self-center" aria-hidden="true"></i>
+    <span class="ml-1 text-nowrap name">{{view.name}}</span>
```
livthomas: Nice usage of `text-overflow: ellipsis`! But you do not really need to define a new class for it since Bootstrap already provides this in `text-truncate` class.

At the same time, it would be nice to add a title on this element in order to show the whole name in case it is truncated. We already do this in other components.

PR #272, opened 2018-02-27
Use collection IDs instead of codes #LMR-690

kubedo8:


src/app/view/view.component.ts
```diff
@@ -113,6 +113,8 @@ export class ViewComponent implements OnInit, OnDestroy {
       perspective: null,
       config: {}
     };
+
+    console.log(this.view);
```
livthomas: This has probably been forgotten here.

src/app/core/search-home/search-home.component.ts
```diff
@@ -77,44 +70,43 @@ export class SearchHomeComponent implements OnInit, OnDestroy {
     }
   }
 
-  public defaultAttribute(document: SearchDocument): string {
-    return document && document.document && document.document.data ? Object.values(document.document.data)[0] : '';
+  public defaultAttribute(document: DocumentModel): string {
+    return document && document.data && document.data ? Object.values(document.data)[0] : '';
```
livthomas: Why do we have `document.data` twice here?

src/app/core/store/collections/collections.effects.ts
```diff
@@ -114,8 +114,8 @@ export class CollectionsEffects {
 
   @Effect()
   public delete$: Observable<Action> = this.actions$.ofType<CollectionsAction.Delete>(CollectionsActionType.DELETE).pipe(
-    switchMap(action => this.collectionService.removeCollection(action.payload.collectionCode)),
-    map(collectionCode => new CollectionsAction.DeleteSuccess({collectionCode: collectionCode})),
+    switchMap(action => this.collectionService.removeCollection(action.payload.collectionId)),
+    map(collectionId => new CollectionsAction.DeleteSuccess({collectionId: collectionId})),
```
livthomas: There is a shortcut for this. Instead of `{collectionId: collectionId}` you can write just `{collectionId}`.

You do not have to change it. I just wanted to let you know about the new ES6 syntax. ;)

src/app/core/store/documents/documents.state.ts
```diff
@@ -50,9 +50,8 @@ export const selectDocumentsByQuery = createSelector(
     documents = DocumentsFilters.filterByQuery(documents, query);
 
     return documents
-      .filter(document => typeof(document) !== 'function')
```
livthomas: +1 for removing these things

src/app/core/store/collections/collections.effects.ts
kubedo8: Yes I know about that, thanks for finding that.

src/app/core/search-home/search-home.component.ts
kubedo8: It was just quick refactor so I forgot it here

PR #273, opened 2018-03-04
Layout registration tab

justmesr:
Unfinished layout


livthomas:
Closing since it is not needed for now.

PR #275, opened 2018-03-05
ui/w/$ORG/$PROJ/view/search/files - handling too long file names #LMR-589

kubedo8:


src/app/shared/search-box/query-item/query-item.component.html
```diff
@@ -8,8 +8,8 @@
 
   <span [class.font-weight-bold]="isCollectionItem()"
         [title]="queryItem.text"
-        class="ml-1 text-truncate">
-    {{queryItem.text}}
+        class="ml-1">
+    {{queryItem.text | truncate:30}}
```
livthomas: Why are we changing it back?

src/app/collection/config/collection-config.component.html
```diff
@@ -2,8 +2,8 @@
      [style.opacity]="collection ? '1' : '0'"
      *ngIf="collection">
   <div class="d-flex flex-row justify-content-start">
-    <a class="fas fa-chevron-left bigger mr-2 text-success" (click)="goToCollectionsPage()"></a>
-    <i class="color-class fa-fw {{ collection.icon }} collection-icon clickable"
+    <i class="fas fa-chevron-left bigger mr-3 clickable" (click)="goToCollectionsPage()"></i>
```
livthomas: Where does this `clickable` class come from and why do we use it everywhere?

src/app/collection/config/collection-config.component.html
```diff
@@ -18,83 +18,88 @@
...
-           i18n-title="@@header.collection.name"
-           title="Edit name of the file">
+    <div class="color-class name flex-grow-2 mr-4 cursor-text"
+         [textContent]="collection.name"
+         (blur)="name.scrollTo(0,0); onNewColectionName(name.textContent)"
+         [style.color]="collection.color"
+         contenteditable="true"
+         placeholder="Collection name"
+         [attr.spellcheck]="false"
+         i18n-title="@@header.collection.name"
```
livthomas: There should be `i18n` attribute for `placeholder` as well.

src/app/core/store/collections/collections.state.ts
```diff
@@ -46,4 +46,8 @@ export const selectCollectionsByQuery = createSelector(selectCollectionsDictiona
   return !query || query.collectionIds.length === 0 ? Object.values(collections) : query.collectionIds.map(id => collections[id]);
 });
 
+export const selectCollection = createSelector(selectCollectionsDictionary, selectWorkspace, (collections, workspace) => {
```
livthomas: `selectCollectionFromWorkspace` would be more clear in my opinion.

src/app/shared/post-it-collections/collection-name/post-it-collection-name.component.ts
```diff
@@ -32,8 +32,9 @@ export class PostItCollectionNameComponent {
 
   @Input() public collection: CollectionModel;
 
-  @Output()
-  public changed = new EventEmitter<string>();
+  @Output() public changed = new EventEmitter<string>();
+  @Output() public selected = new EventEmitter();
+  @Output() public unselected = new EventEmitter();
```
livthomas: Cannot the last two be merged into a single output with `boolean` parameter? We should always try to reduce the number of interactions between components.

The other thing is that JS events are usually called `change`, `select`, etc. But it is not really important.

src/app/shared/search-box/query-item/query-item.component.html
kubedo8: Because of long collection names. For my opinion 30 characters is enough...

src/app/collection/config/collection-config.component.html
kubedo8: It is defined in `common.scss` and has our default behaviour of clickable items (green colour with darker green on hover)

src/app/collection/config/collection-config.component.html
kubedo8: Yes..thanks

src/app/core/store/collections/collections.state.ts
kubedo8: Good idea.

src/app/shared/search-box/query-item/query-item.component.html
livthomas: Oh, I see. We changed it to `text-truncate` but forgot to set the maximum width.

PR #277, opened 2018-03-12
Organization settings & User tab

kubedo8:


src/app/shared/input/filter-box/filter-box.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  @Input() public value: string;
+
+  @Input() public placeholder: string;
+
+  @Output() public textChange: EventEmitter<string> = new EventEmitter();
+
+  @Output() public textSubmit: EventEmitter<string> = new EventEmitter();
+
+  public mPlaceholder: string;
+  public mCurrentValue: string;
```
livthomas: I'm just curious, why `m` prefix?

src/app/shared/utils/validator.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export class Validator {
```
livthomas: We already have validators in the core module. So it should be moved there and rename accordingly.

src/app/view/view-controls/share-dialog/share-dialog.component.ts
```diff
@@ -17,35 +17,54 @@
...
   public constructor(private notificationService: NotificationService,
-                     private userService: UserService) {
+                     private store: Store<AppState>) {
   }
 
   public ngOnInit() {
-    this.userService.getUsers().subscribe((users: User[]) => {
-      this.users = users;
-    });
+    this.subscribeData();
```
livthomas: +1 on making this reactive.

src/app/workspace/organization/organization-settings.component.html
```diff
@@ -1,98 +1,103 @@
...
-          <icon-picker class="dropdown-item" [(icon)]="organization.icon"
-                       (selectedChange)="updateOrganization()"></icon-picker>
-        </div>
-
-        <input class="color-class name edit-border" [(ngModel)]="organization.name" (blur)="name.valid && updateOrganizationName()"
-               name="name"
-               [style.color]="organization.color" i18n-title="@@header.organization.name"
-               title="Edit name of the organization"
-               #name="ngModel" required minlength="3" maxlength="24">
+<div class="col-12 col-lg-11 col-xl-10 m-auto" *ngIf="organization">
```
livthomas: This component is just huge. We should try to split it into the smaller ones in the future.

src/app/core/dto/group.ts
```diff
@@ -19,8 +19,7 @@
 
 export interface Group {
 
-  id?: string;
+  id: string;
```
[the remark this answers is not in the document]
kubedo8: We create new group with `GroupModel` interface. This is dto class and it should be alwas defined.

src/app/shared/utils/validator.ts
kubedo8: Okay

src/app/core/store/users/users.state.ts
```diff
@@ -21,14 +21,38 @@ import {createEntityAdapter, EntityState} from '@ngrx/entity';
...
 import {AppState} from '../app.state';
 import {UserModel} from './user.model';
+import {selectOrganizationByWorkspace} from "../organizations/organizations.state";
+import {selectAllGroups} from "../groups/groups.state";
+import {UserFilters} from "./user.filters";
+import {GroupModel} from "../groups/group.model";
 
 export interface UsersState extends EntityState<UserModel> {
 
+  filter: string
```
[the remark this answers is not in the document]
kubedo8: When user select another tab and return to `users` tab, filter is removed. I think that this behaviour is not very user friendly. Do you have any other solution?

src/app/shared/input/filter-box/filter-box.component.ts
kubedo8: I wanted to distinguish between `input` variables and `member` variables.

src/app/workspace/organization/organization-settings.component.html
kubedo8: Do you think that making `resource-header` is good idea? I was thinking about that, but there are some differences so I am not sure about that.. But of course I can split this component into the smaller ones..

src/app/core/store/organizations/organizations.effects.ts
```diff
@@ -74,12 +77,26 @@ export class OrganizationsEffects {
...
+
+      const paramMap = RouteFinder.getFirstChildRouteWithParams(this.router.routerState.root.snapshot).paramMap;
+      const orgCodeInRoute = paramMap.get('organizationCode');
+
+      if (orgCodeInRoute && orgCodeInRoute === oldOrganization.code && organization.code !== oldOrganization.code) {
+        const paths = this.router.routerState.snapshot.url.split("/").filter(path => path);
+        const index = paths.indexOf(oldOrganization.code, 1);
+        if(index !== -1){
+          paths[index] = organization.code;
+          actions.push(new RouterAction.Go({path: paths}))
```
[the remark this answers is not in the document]
kubedo8: Okay

src/app/core/dto/group.ts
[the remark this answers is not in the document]
livthomas: I mean when you create a new group object and send it to the server for the first time, it does not have `id` field defined, right?

src/app/core/store/users/users.state.ts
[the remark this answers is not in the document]
livthomas: I see your point. In that case I would recommend to pass such information using router's query params and remove it from them in `OnDestroy` method of the parent component.

src/app/workspace/organization/organization-settings.component.html
livthomas: I have not really analyzed it into the depth. If we create two components that are similar we can then work further and extract them into the new abstract one.

PR #280, opened 2018-03-17
Create new table header #LMR-742

livthomas:


kubedo8:
Really nice work!

It's difficult to understand how it works, so I don't write any comments.

Maybe one question:  We have two table perspectives until you finish `table2`?

livthomas:
I don't want to remove the old table perspective unless you are not able to do the same things with the new one. Hopefully, it will be finished this week and we can remove it then.

When it comes to how it works, it's not that difficult. Basically, you have a set of tables in the store so you can work with multiple tables on a single screen. The table takes its structure (parts, column layout, etc.) from `tables` store and the data (collections, links, documents) from other stores and react to changes to both of them. If you want to change the structure of the table, you just need to dispatch some table action and the components will re-render. The table parts are created based on the query so in order to change them (switch, remove, etc.), you need to change the query and the table will change accordingly.

kubedo8:
Thanks for the explanation.

Maybe we can add some description to large pull requests like this for better understanding what's going on.

PR #283, opened 2018-03-21
LMR-698_2

anujgoel89:
Icon Picker Visibility change so it is always on top of other elements


marvec:
Not going to merge it for now.

PR #286, opened 2018-03-22
Add chart perspective

justmesr:


src/app/view/perspectives/chart/model/collection-attribute-pair-packer.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import {DocumentModel} from '../../../../core/store/documents/document.model';
+import {CollectionAttributePair} from './collection-attribute-pair';
+import {CollectionModel} from '../../../../core/store/collections/collection.model';
+
+type Pair = [CollectionModel, string];
+
+export class Packer {
+
+  public pack(documents: DocumentModel[]): CollectionAttributePair[] {
```
livthomas: This is exactly the Java style we discussed today. It's not a bad solution. But it would be much cleaner to move this method to `collection-attribute-pair.ts` and have it there as function called for example `createCollectionAttributePair`. The only exported function would be this one and the other private ones would not be exported and so they will not be visible from other modules.

src/app/view/perspectives/chart/sorter/creation-date-sorter.ts
```diff
@@ -0,0 +1,40 @@
...
+        return 1;
+      });
+
+    } else {
+      return [];
+    }
+  }
+
+  private copy(documents: DocumentModel[]): DocumentModel[] {
+    return documents.slice();
```
livthomas: Just wondering. Why do we need this method here? You can just call `documents.slice()` in the method above.

CONTRIBUTING.md
```diff
@@ -126,6 +126,81 @@ Look at the changes (git diff) and apply them to other languages (such as `messa
 Preferably, you should also provide a translation to these languages by adding `<target>` element to each new translation unit.
 If you do not speak those languages, ask maintainers for a translation when you send a pull-request.
 
+### Adding NPM packages
+
+```
+npm install <package name> --save
+```
+Then add the package to `.angular-cli.json` under `"scripts"` key to include it in Angular CLI compilation:
```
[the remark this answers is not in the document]
justmesr: It's not necessary, but it was in the tutorial for adding Plotly to Angular.. and I guess it won't do any harm. (I can remove it though).

src/app/view/perspectives/chart/chart-perspective.component.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  set documents(value) {
+    this._documents = value;
+    this.documentAttributes = this.usableAttributes();
+  }
+
+  public documentAttributes: CollectionAttributePair[];
+
+  private subscriptions: Subscription;
```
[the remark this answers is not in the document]
justmesr: I want it there.. just in case someone changes the intial subscription and forgets to add it in `ngOnDestroy`. Nice find though. :)

src/app/view/perspectives/chart/chart-perspective.component.ts
```diff
@@ -0,0 +1,133 @@
...
+  private _documents: DocumentModel[];
+
+  get documents() {
+    return this._documents;
+  }
+
+  set documents(value) {
+    this._documents = value;
+    this.documentAttributes = this.usableAttributes();
+  }
```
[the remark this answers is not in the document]
justmesr: Complicated it for the sake of declarativness. Instead of adding the attribute recalculation next to every documents assignment (even though there is currently only one), just telling it to recalculate every time, the documents change.

CONTRIBUTING.md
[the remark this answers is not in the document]
livthomas: Ok, keep it here since we want to move to Angular CLI in the future.

justmesr:
@livthomas

package.json
```diff
@@ -44,6 +44,7 @@
     "@ngrx/store": "^5.2.0",
     "@ngrx/store-devtools": "^5.2.0",
     "@ngx-translate/i18n-polyfill": "^0.1.2",
+    "@types/plotly.js": "^1.35.0",
```
livthomas: Types should go to `devDependencies`. We do not want to bundle them with the application.

src/app/view/perspectives/chart/chart-perspective.component.html
```diff
@@ -34,14 +34,14 @@
...
         (attributeSelected)="attributeX = $event">
       </attribute-select>
     </div>
 
   </div>
 </ng-container>
 
-<ng-template #loading>
+<ng-template #loading i18n-title="@@chart.loading">
   Loading...
```
livthomas: I am not sure how this will behave. I would recommend to create inner `<span>` element with `i18n` attribute.

src/app/view/perspectives/chart/chart-perspective.component.html
```diff
@@ -18,7 +18,7 @@
         [@slide]
         [attributes]="documentAttributes"
         [preSelected]="attributeY"
-        [emptyDescription]="'Y Axis'"
+        [emptyDescription]="'Y ' + axisTitle()"
```
livthomas: Maybe I would translate the whole thing separately for each axis as it might be translated differently in various languages.

PR #287, opened 2018-03-22
Fix inability to create attributes

justmesr:
It was impossible to create new attributes in collection config.

### DIFF
--- a/src/app/collection/config/tab/attribute-list/attribute-list.component.ts
+++ b/src/app/collection/config/tab/attribute-list/attribute-list.component.ts
@@ -26,6 +26,7 @@ import * as Const from '../constraints';
 
 import {ConfiguredAttribute} from './configured-attribute';
 import {ConstraintSuggestion} from './constraint-suggestion';
+import {Attribute} from '../../../../core/dto';
 
 @Component({
   selector: 'attribute-list',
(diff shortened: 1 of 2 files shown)


src/app/collection/config/tab/attribute-list/attribute-list.component.ts
```diff
@@ -71,12 +72,11 @@ export class AttributeListComponent {
               private notificationService: NotificationService) {
   }
 
-  private emptyAttribute(): ConfiguredAttribute {
+  private emptyAttribute(): Attribute {
```
livthomas: We should rather use `AttributeModel` class so we do not deal with DTO objects outside of the services and store.

src/app/collection/config/tab/attribute-list/attribute-list.component.ts
```diff
@@ -191,7 +191,7 @@ export class AttributeListComponent {
 
     this.suggestions = Const.constraints.map(flattenConstraints)
       .reduce(flatten, [])
-      .filter(suggestion => startsWith(suggestion.name, this.activeAttribute.newConstraint))
+      .filter(suggestion => startsWith(suggestion.name, this.activeAttribute['newConstraint']))
```
livthomas: Why do we need to keep this `newConstraint` field on an attribute? In general, we should not modify existing objects since they can come from the store and we would like to be able to access older versions of data in order to make application behavior predictable and debugging much easier.

justmesr:
Yes the component isn't using NgRx yet. The `newConstraint` field was an unfortunate solution from my side. However since it's not using Store,  it's not causing harm.

PR #290, opened 2018-03-22
 search/records - no data are loaded at first, then cannot unfold details for individual records #LMR-782

kubedo8:
…ils for individual records #LMR-782

### DIFF
--- a/src/app/core/rest/search.service.ts
+++ b/src/app/core/rest/search.service.ts
@@ -46,7 +46,7 @@ export class SearchService {
               private homePageService: HomePageService) {
     this.store.select(selectWorkspace)
       .pipe(
-        filter(workspace => !isNullOrUndefined(workspace) && !isNullOrUndefined(workspace.organizationCode) && !isNullOrUndefined(workspace.projectCode))
+        filter(workspace => !!workspace && !!workspace.organizationCode && !!workspace.projectCode)
       )
       .subscribe(workspace => this.workspace = workspace);
   }
--- a/src/app/core/rest/search.service.ts
+++ b/src/app/core/rest/search.service.ts
@@ -29,7 +29,7 @@ import {Document} from '../dto/document';
 import {View} from '../dto/view';
 import {SuggestionType} from '../dto/suggestion-type';
 import {Observable} from 'rxjs/Observable';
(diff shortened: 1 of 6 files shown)


src/app/core/rest/search.service.ts
```diff
@@ -44,7 +44,11 @@ export class SearchService {
   constructor(private http: HttpClient,
               private store: Store<AppState>,
               private homePageService: HomePageService) {
-    this.store.select(selectWorkspace).subscribe(workspace => this.workspace = workspace);
+    this.store.select(selectWorkspace)
+      .pipe(
+        filter(workspace => !isNullOrUndefined(workspace) && !isNullOrUndefined(workspace.organizationCode) && !isNullOrUndefined(workspace.projectCode))
```
livthomas: I'm just curious. Why do we need to use `isNullOrUndefined` here?

src/app/shared/utils/array.utils.ts
```diff
@@ -50,3 +50,9 @@ export function getArrayDifference<T>(bigArray: T[], smallArray: T[]): T[] {
 
   return bigArray.filter(item => !smallArray.includes(item));
 }
+
+export function arrayIntersection(array1: string[], array2: string[]): string[] {
```
livthomas: This method should be made generic in order to be reusable. Look at the method a few lines above.

src/app/core/rest/search.service.ts
kubedo8: I wanted to use `workspace && workspace.organizationCode && workspace.projectCode`, but it return string. Do you have any other idea?

src/app/shared/utils/array.utils.ts
kubedo8: Oh yes :)

src/app/core/rest/search.service.ts
livthomas: The typical trick in JavaScript is:
`!!workspace && !!workspace.organizationCode && !!workspace.projectCode`

justmesr: Or `Boolean(workspace && workspace.organizationCode && workspace.projectCode)`, should work the same.

PR #296, opened 2018-03-23
Wrong number of records and attributes shown on the file card #LMR-755

kubedo8:


src/app/core/store/documents/documents.effects.ts
```diff
@@ -110,12 +118,18 @@ export class DocumentsEffects {
...
-
       return this.documentService.updateDocument(documentDto).pipe(
         map(dto => DocumentConverter.fromDto(dto))
       );
     }),
-    map(document => new DocumentsAction.UpdateDataSuccess({document: document})),
+    withLatestFrom(this.store$),
+    flatMap(([document, state]) => {
+      const collection = state.collections.entities[document.collectionId];
+      const oldDocument = state.documents.entities[document.id];
```
livthomas: We should access store through selectors and not directly. Otherwise, we will need to change our code at many places when we make a small change in its structure.

src/app/shared/utils/array.utils.ts
```diff
@@ -54,5 +50,5 @@ export function getArrayDifference<T>(bigArray: T[], smallArray: T[]): T[] {
 export function arrayIntersection<T>(array1: T[], array2: T[]): T[] {
   const a = array1 || [];
   const b = array2 || [];
-  return a.filter(x => b.indexOf(x) !== -1);
+  return a.filter(x => b.includes(x));
```
livthomas: +1

src/app/core/store/documents/documents.effects.ts
```diff
@@ -155,4 +178,63 @@ export class DocumentsEffects {
...
+    this.decrementAttributesInCollection(collectionCopy, removedAttributes);
+
+    return new CollectionsAction.UpdateSuccess({collection: collectionCopy});
+  }
+
+  private createRemoveCollectionsAttributesAction(collection: CollectionModel, document: DocumentModel): Action {
+    const collectionCopy = {...collection, documentsCount: Math.max(collection.documentsCount - 1, 0)};
+
+    const deletedAttributes = Object.keys(document.data);
+    this.decrementAttributesInCollection(collectionCopy, deletedAttributes);
```
livthomas: Not important, just code style. Much cleaner solution would be to return a list of attributes from the function and not modify its parameters.

src/app/core/store/documents/documents.effects.ts
```diff
@@ -155,4 +178,63 @@ export class DocumentsEffects {
...
+
+    return new CollectionsAction.UpdateSuccess({collection: collectionCopy});
+  }
+
+  private decrementAttributesInCollection(collection: CollectionModel, attributeIds: string[]) {
+    collection.attributes = collection.attributes.map(attribute => {
+        if (attributeIds.includes(attribute.id)) {
+          attribute.usageCount = Math.max(attribute.usageCount - 1, 0);
```
livthomas: You are modifying the object in the store here!
```typescript
return {...attribute, usageCount: Math.max(attribute.usageCount - 1, 0)};
```

src/app/core/store/documents/documents.effects.ts
```diff
@@ -155,4 +181,59 @@ export class DocumentsEffects {
...
+          if (attributeIndex !== -1) {
+            attributeToInc.splice(attributeIndex, 1);
+            return {...attribute, usageCount: attribute.usageCount + 1};
+          }
+        }
+        return attribute;
+      }
+    );
+
+    attributeToInc.map(attributeId => ({id: attributeId, name: attributeId, constraints: [], usageCount: 1}))
```
livthomas: Be aware that `name` does not always equal `id`. I would recommend using `extractAttributeName` from attribute utils here. After the attribute model refactoring there will no longer be this problem. But for now, the current behavior may break the table.

PR #300, opened 2018-03-23
Creating duplicate organisation name/code #LMR-784

kubedo8:


src/app/core/store/organizations/organizations.effects.ts
```diff
@@ -50,6 +50,18 @@ export class OrganizationsEffects {
...
 
+  @Effect()
+  public getCodes$: Observable<Action> = this.actions$.ofType<OrganizationsAction.GetCodes>(OrganizationsActionType.GET_CODES).pipe(
+    switchMap(() => this.organizationService.getOrganizationsCodes()),
+    map((organizationCodes) => new OrganizationsAction.GetCodesSuccess({organizationCodes})),
+    catchError((error) => Observable.of(new OrganizationsAction.GetCodesFailure({error: error})))
+  );
+
+  @Effect({dispatch: false})
+  public getCodes$Failure$: Observable<Action> = this.actions$.ofType<OrganizationsAction.GetCodesFailure>(OrganizationsActionType.GET_CODES_FAILURE).pipe(
```
livthomas: Too many dollars. :dollar:

src/app/workspace/workspace-chooser/resource-chooser/resource-chooser.component.scss
```diff
@@ -138,7 +138,7 @@ $bookmark-border-color: #bdbdbd;
             }
 
             .lmr-warning {
-              border: 1px solid orange !important;
+              border: 1px solid red !important;
```
livthomas: What about using `.border-danger` from Bootstrap instead?

kubedo8:
And what do you recommend? I think it is same as other things we do through the store. I'm not happy with that, but we need these codes immediately after each operation.

livthomas:
Maybe we can always get them from the server as you start typing.

kubedo8:
But it's not guaranteed they will come in time.

PR #303, opened 2018-03-24
#LMR-702

justmesr:


src/app/shared/post-it-collections/project-getter.ts
```diff
@@ -0,0 +1,68 @@
...
+import {ProjectModel} from '../../core/store/projects/project.model';
+import {AppState} from '../../core/store/app.state';
+import {Store} from '@ngrx/store';
+import {OrganizationsAction} from '../../core/store/organizations/organizations.action';
+import {selectOrganizationByCode} from '../../core/store/organizations/organizations.state';
+import {ProjectsAction} from '../../core/store/projects/projects.action';
+import {selectProjectByCode} from '../../core/store/projects/projects.state';
+import {OrganizationModel} from '../../core/store/organizations/organization.model';
+
+export class ProjectGetter {
```
kubedo8: We have always current organization and project in the store (look at `workspace.guard.ts`). So I think this class is needless.

src/app/shared/utils/layout/post-it-layout.ts
```diff
@@ -24,7 +24,7 @@ export class PostItLayout {
 
   protected layout: any;
 
-  protected insertingElementsAtIndex: number = 2;
+  protected insertingElementsAtIndex: number = 0;
```
kubedo8: Are you sure with that? I recently changed it, because it adds post-its to end of layout and I think this is not very good behaviour.

src/app/view/perspectives/search/all/search-all.component.ts
```diff
@@ -48,42 +48,50 @@ export class SearchAllComponent implements OnInit, OnDestroy {
...
+
+  private subscribeOnQuery() {
+    return this.store.select(selectNavigation).pipe(
+      filter((navigation: NavigationState) => {
+        return Boolean(navigation && navigation.workspace && navigation.workspace.organizationCode && navigation.workspace.projectCode);
+      }),
+      map(navigation => navigation.query),
+      filter(query => !isNullOrUndefined(query)),
+      map(query => ({...query, page: 0, pageSize: 100})), // TODO implement pagination logic
+      tap(query => this.store.dispatch(new CollectionsAction.Get({query}))),
```
kubedo8: Collections are always queried before the page is loaded. So we don't need to load them again.

src/app/shared/post-it-collections/project-getter.ts
justmesr: Hm.. I tried selecting them directly, without dispatching `GET`, but the subscribtion callback was never called.

I can try it out again later. I migh have tried wrong selectors (selectedProject didn't work for sure).

src/app/shared/utils/layout/post-it-layout.ts
justmesr: Well, it specifies the index the new elements should be added at.
It was 2, because add and import were indexes 0, and 1. Now they
are replaced by buttons, so 0 is similiar to inserting at the start.

Didn't it work now as expected, since you're asking?

kubedo8: It's okay then.

src/app/shared/post-it-collections/project-getter.ts
kubedo8: Try to create new selector `selectCurrentOrganization` and `selectCurrentProject`. `selectedProject` works right only for workspace.

justmesr: Okay, thanks.

src/app/shared/post-it-collections/add-button/post-it-collection-add-button.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  @Output()
+  public clicked = new EventEmitter<CollectionModel>();
+
+  @Input()
+  public disabled: boolean;
+
+  public add() {
+    const newCollection = {
+      ...this.emptyCollection(),
+      correlationId: CorrelationIdGenerator.generate()
```
livthomas: It's totally up to you but much cleaner design would to have a generic add button, emit empty `click` event and react to it in the component where it is used.

src/app/shared/post-it-collections/import-button/post-it-collection-import-button.component.scss
```diff
@@ -0,0 +1 @@
+@import "../../variables";
```
livthomas: We do not need empty SCSS files.

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -290,4 +245,12 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
     return HashCodeGenerator.hashString(collection.id || collection.correlationId);
   }
 
+  public notifyOfError(message: string) {
+    this.store.dispatch(new NotificationsAction.Error({message}));
```
livthomas: You do not need to show notifications by dispatching store actions. You can do so directly. It is not a problem, you are just going around for no good reason. We have actions for them to be able to easily switch some error action in store to notification action.

src/app/shared/post-it-collections/project-getter.ts
livthomas: I don't see much added value in this class at all. You just need to create the right selector in the store and it will be a one-liner to get the current project in a component.

src/app/shared/post-it-collections/import-button/post-it-collection-import-button.component.html
```diff
@@ -0,0 +1,24 @@
+<button type="button"
+        class="btn btn-outline-{{ disabled ? 'secondary' : 'primary' }}"
```
livthomas: Why do we use `primary` color here? It does not fit to our design at all. It should be `success`.

The other thing is that the button should probably not be shown when the user is not able to click it.

src/app/view/perspectives/search/all/search-all.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="mt-2">
-  <ng-container *ngIf="loading()">
+  <ng-container *ngIf="loading()" i18n="@@search.all.loading">
     Loading...
```
livthomas: The same as before. I am afraid that `<ng-container>` is not real HTML element and so it will not be in the DOM. We should not add attributes on it.

src/app/view/perspectives/search/all/search-all.component.ts
```diff
@@ -48,42 +48,50 @@ export class SearchAllComponent implements OnInit, OnDestroy {
...
-      ).subscribe();
-
-    this.collectionsSubscription = this.store.select(selectCollectionsByQuery).subscribe(collections => {
-      this.collections = collections;
+    [
+      this.subscribeOnQuery(),
+      this.subscribeOnCollection(),
+      this.subscribeOnDocuments(),
+      this.subscribeOnViews()
+    ].forEach(subscription => this.subscriptions.add(subscription));
```
livthomas: Nice!

src/app/view/perspectives/search/all/search-all.component.ts
livthomas: And it is the same for views after the latest changes.

PR #304, opened 2018-03-24
Fix inability to create collections

SimonKocurek:


src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -245,13 +245,9 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
...
     const collectionCopy = {...collection, name: newName};
 
-    if (!collection.name) {
-      return;
-    }
-
     if (collection.id) {
       this.updateCollection(collectionCopy);
-    } else {
+    } else if (newName) {
```
livthomas: Should not there be a condition here so it will not allow to use empty collection name? Or is it checked somewhere else?

PR #305, opened 2018-03-24
Fix aborting REST API requests #LMR-795

livthomas:
We should never use `switchMap` operator when calling remote services. See [this article](https://blog.angularindepth.com/switchmap-bugs-b6de69155524) for more details.


kubedo8:
What a good catch! Thanks for the article 👍

PR #306, opened 2018-03-26
Fix Redux DevTools freezing #LMR-797

livthomas:
Have you ever wondered why Redux DevTools are often freezing? Here is the answer:

Angular router classes create a complicated hierarchies with many cyclic references. Redux DevTools always try to serialize the content of the store and they go into the depth. The result was 2GB RAM taken by Redux DevTools and the eventual crash of the extension.

See [this NgRx issue](https://github.com/ngrx/platform/issues/97) and [this section of the documention](https://github.com/ngrx/platform/blob/master/docs/router-store/api.md#custom-router-state-serializer).

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -19,11 +19,12 @@
 
 import {NgModule} from '@angular/core';
 import {RouterModule, Routes} from '@angular/router';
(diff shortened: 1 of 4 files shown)


kubedo8:
Awesome! I have prayed many times for it to work well..many thanks 👍

PR #309, opened 2018-04-13
LMR-786 running build/start trough angular cli

karelhala:
### Loading of variables
When build or start is executed it will first read env variables which we defined earlier and stores them in file `src/environment/.env.json` and in `.env` file in root folder

#### `src/environment/.env.json`
This file is then loaded in both production and development environments. From now on if you want to load variable you should load it like this (based on your location)
```typescript
import { environment } from '../environments/environment';
```
Which then can be used as `environment.MY_VARIABLE` where `MY_VARIABLE` is OS env variable.

#### `.env`
This file contains variable `LUMEER_SWITCHES` which has all parameters for angular cli, do not change this file (it's auto generated) and do not set `LUMEER_SWITCHES` it will not take any effect. When build/start is running first we'll load variables from this file and then pass them to ng-cli.

There are two possible variations:
* defaultSwitches - output path and base href is set via this settings
* advancedSwitches - when AOT is set use translations as well

### General changes
* scss now loads trough angular cli
* output path passed as parameter from `.env` file
* base href passed as parameter from `.env` file
* variables for translation passed as parameter from `.env` file
### Drawbacks
* we are no longer able to pass variables to index.html during build, but have to add some work around using javascript. Not that big of a deal


karelhala:
@livthomas I was able to test translations for start, however I don't know how to test if build is correct. I also kept old config files there, so if you want I can add build/start trough old config if something fails.

karelhala:
@livthomas I forgot to post my comment earlier.... The problem with quill editor is that types for quill edirot are missing some functionalities (like `import`) so I removed them in order to pass the build. I've refactored the loading of default values for variables as well, however it's bound to bash due to env variables. Since Windows is using bash as well we're good to go. The biggest challenge was that we expect some varibales for build to be set so I had to do this default variable set in file `config/env-vars.js`.

To run AOT you just simply run
```bash
$ AOT=true npm run start
```

To run start with czech translation run
```bash
$ AOT=true I18N_LOCALE=cs npm run start
```

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "1.0.0",
   "description": "Lumeer Web UI",
   "main": "index.js",
+  "config": {
+    "ENV_VAR": "echo $bla"
```
livthomas: Do we need this or did you just use it for testing?

src/typings.d.ts
```diff
@@ -3,3 +3,8 @@ declare var module: NodeModule;
 interface NodeModule {
   id: string;
 }
+
+declare module '*.json' {
+  const value: any;
+  export default value;
```
livthomas: What is this module good for?

livthomas:
I have also noticed that `raw-loader` used by Angular CLI tries to include files from `target` directory created by Maven and it runs out of memory when it tries to process `lumeer-web.war` which is quite a big file. See the following output where `...` means 11MB of symbols like those above and under them:

```
WARNING in ./node_modules/raw-loader!./target/lumeer-web.war
Module parse failed: Unterminated string constant (1:17)
You may need an appropriate loader to handle this file type.
| module.exports = "PK\u0003\u0004\n\u0000\u0000\u0000\u0000\u0000\u001a]_K\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\t\u0000\u0000\u0000META-INF/PK\u0003\u00
...
...
...
\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000��oRH\u0000META-INF/maven/io.lumeer/web-ui/pom.propertiesPK\u0005\u0006\u0000\u0000\u0000\u00002\u00002\u0000�\u000e\u0000\u0000)SH\u0000\u0000\u0000"

webpack: Compiled with warnings.
webpack: Compiling...
 95% emitting                                                                      
<--- Last few GCs --->

[3805:0x33fe790]   149744 ms: Mark-sweep 1287.9 (1430.4) -> 1287.8 (1429.9) MB, 848.3 / 0.4 ms  allocation failure GC in old space requested
[3805:0x33fe790]   150601 ms: Mark-sweep 1287.8 (1429.9) -> 1287.8 (1384.9) MB, 857.5 / 0.4 ms  last resort GC in old space requested
[3805:0x33fe790]   151383 ms: Mark-sweep 1287.8 (1384.9) -> 1287.8 (1374.9) MB, 781.9 / 0.4 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x20e90cfa57c1 <JSObject>
    1: fromString(aka fromString) [buffer.js:314] [bytecode=0x288fb4d7ce09 offset=164](this=0x1e8d099022d1 <undefined>,string=0x312089fa9a71 <Very long string[69687412]>,encoding=0x20e90cfb4a51 <String[4]: utf8>)
    2: from [buffer.js:177] [bytecode=0x288fb4d7c911 offset=11](this=0x11c7c84b6401 <JSFunction Buffer (sfi = 0x20e90cffaf31)>,value=0x312089fa9a71 <Very long string[69687412]>,encoding...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [ng]
 2: 0x11e7fec [ng]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [ng]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [ng]
 5: v8::internal::Factory::NewRawTwoByteString(int, v8::internal::PretenureFlag) [ng]
 6: v8::internal::String::SlowFlatten(v8::internal::Handle<v8::internal::ConsString>, v8::internal::PretenureFlag) [ng]
 7: v8::internal::String::Flatten(v8::internal::Handle<v8::internal::String>, v8::internal::PretenureFlag) [ng]
 8: v8::String::WriteUtf8(char*, int, int*, int) const [ng]
 9: node::StringBytes::Write(v8::Isolate*, char*, unsigned long, v8::Local<v8::Value>, node::encoding, int*) [ng]
10: node::Buffer::New(v8::Isolate*, v8::Local<v8::String>, node::encoding) [ng]
11: 0x120ce14 [ng]
12: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) [ng]
13: 0xb7a9ac [ng]
14: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*) [ng]
15: 0x2ae32fd842fd
```

livthomas:
@karelhala I have fixed wrong proxy configuration. We still need to figure out how to ignore Maven `target` folder.

package.json
karelhala: It was for testing, I'll remove it.

src/typings.d.ts
karelhala: It's for typescript to read json files and split them into correct values.

livthomas:
I have already migrated the application to use Angular CLI in the other PR (#374). The configuration has changed a lot in Angular CLI 6 so I have decided to do this change together with Angular 6 update.

@karelhala But I have reused your solution for environment variables. Thank you very much for this!

PR #310, opened 2018-04-13
added contact form #LMR-806

marvec:


src/app/core/store/organizations/contact/contacts.effects.ts
```diff
@@ -0,0 +1,78 @@
...
+import {Observable} from "rxjs/Observable";
+import {I18n} from "@ngx-translate/i18n-polyfill";
+import {ContactsAction, ContactsActionType} from "./contacts.action";
+
+@Injectable()
+export class ContactsEffects {
+  @Effect()
+  public getContact$: Observable<Action> = this.actions$.pipe(
+    ofType<ContactsAction.GetContact>(ContactsActionType.GET_CONTACT),
+    mergeMap(action => this.organizationService.getOrganizationContact(action.payload.organizationId)),
```
kubedo8: Method assumes `code` not `id`.

src/app/core/store/organizations/contact/contacts.effects.ts
```diff
@@ -0,0 +1,78 @@
...
+    map(() => {
+      const message = this.i18n({id: 'organization.contact.get.fail', value: 'Cannot read contact information'});
+      return new NotificationsAction.Error({message});
+    })
+  );
+
+  @Effect()
+  public setContact$: Observable<Action> = this.actions$.pipe(
+    ofType<ContactsAction.SetContact>(ContactsActionType.SET_CONTACT),
+    mergeMap(action => this.organizationService.setOrganizationContact(action.payload.organizationId, ContactConverter.toDto(action.payload.contact))),
```
kubedo8: Same here..

src/app/workspace/organization/detail/organization-detail.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  ngOnInit(): void {
+    this.subscribeToStore();
+    this.requestData();
+  }
+
+  private subscribeToStore() {
+    this.contactSubscription = this.store.select(selectContactByCode(this.organizationSettingsComponent.organization.code))
+      .subscribe(contact => this.contactForm.setContact(contact));
```
kubedo8: We use observable as input property into component. You can see it `WorkspaceChooserComponent for example.

src/app/workspace/organization/detail/organization-detail.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  @ViewChild('contactForm')
+  private contactForm: ContactFormComponent;
+
+  private contactSubscription: Subscription;
+
+  constructor(private i18n: I18n,
+              private router: Router,
+              private store: Store<AppState>,
+              private notificationService: NotificationService,
+              private organizationSettingsComponent: OrganizationSettingsComponent) {
```
kubedo8: I'm not sure about that. You bind this component to its parent and I think it's not good idea.

src/app/core/store/organizations/contact/contacts.state.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const contactsAdapter = createEntityAdapter<ContactModel>({selectId: contact => contact.code});
+
+export const initialContactsState: ContactsState = contactsAdapter.getInitialState({
+});
+
+export const selectContactsState = (state: AppState) => state.contacts;
+export const selectAllContacts = createSelector(selectContactsState, contactsAdapter.getSelectors().selectAll);
+export const selectContactByCode = (code) => createSelector(selectAllContacts, contacts => {
+  return contacts.find(contact => contact.code === code);
```
kubedo8: I'm confused, because you set `organizationCode` here as parameter, but you use `contact.code` to find Contact. It means that contact code is made of organization code?

src/app/core/store/organizations/contact/contacts.effects.ts
marvec: Sure, this is just a wrong name of the attribute in the object. It is a copy paste error, I must fix that!

src/app/core/store/organizations/contact/contacts.state.ts
marvec: Contact code is the code of organization for which the contact is. This is the binding attribute.

kubedo8: I think we should use `id` instead of `code`, because `code` is not immutable attribute.

PR #314, opened 2018-04-21
Handling payments in UI LMR-815

marvec:


src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.html
```diff
@@ -2,30 +2,34 @@
...
-          <td>2018-02-20</td>
-          <td><span class="badge badge-pill badge-success p-2">Business</span></td>
-          <td>10</td>
-          <td>€402.00</td>
-          <td>2019-02-19</td>
+        <tr *ngFor="let payment of payments">
+          <td>{{ payment.date | date : 'shortDate'}}</td>
+          <td>
+            <span *ngIf="payment.serviceLevel == 'FREE'" class="badge badge-pill badge-danger p-2" i18n="@@organizations.tab.detail.payments.trial">Trial</span>
+            <span *ngIf="payment.serviceLevel != 'FREE'" class="badge badge-pill badge-success p-2" i18n="@@organizations.tab.detail.payments.business">Business</span>
```
livthomas: It is not a problem in this particular case but we should always use triple equals (`===` and `!==`) in JavaScript/TypeScript otherwise we might end up with really strange behavior of the conditions. See [this chapter](http://speakingjs.com/es5/ch09.html#equality_operators) of Speaking JavaScript book.

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.html
```diff
@@ -2,30 +2,34 @@
...
+          </td>
+          <td>{{ payment.users }}</td>
+          <td>{{ payment.currency }} {{ payment.amount | number:'1.2-2' }}</td>
+          <td *ngIf="payment.state == 'PAID'" i18n="@@organizations.tab.detail.payments.paid">Paid</td>
+          <td *ngIf="payment.state != 'PAID'" i18n="@@organizations.tab.detail.payments.ordered">Ordered <button type="button" class="btn btn-sm btn-success" (click)="refreshPayment(payment.paymentId)"><i class="fas fa-sync-alt"></i></button></td>
```
livthomas: Isn't it better to place a translation just around the `Ordered` text?

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.ts
```diff
@@ -17,18 +17,85 @@
...
+export class PaymentsListComponent implements OnInit, OnDestroy {
 
-  constructor() { }
+  private organization: OrganizationModel;
+  private organizationSubscription: Subscription;
+
+  public payments: PaymentModel[];
+  private paymentsSubscription: Subscription;
+
+  private getPaymentSubscription: Subscription;
```
livthomas: If you have multiple subscriptions in the component and the only place where you unsubscribe from them is `ngOnDestroy` method, you can combine them in a single subscription.

The field definition would look like this:

```typescript
private subscriptions = new Subscription();
```

Payment subscription (or any other subscription) would look like this:
```typescript
this.subscriptions.add(
  this.store.select(selectPaymentsByOrganizationIdSorted(this.organization.id))
      .pipe(filter(payments => !isNullOrUndefined(payments) && payments.length > 0))
      .subscribe(payments => this.payments = payments)
);
```

And `ngOnDestroy` method would look like this:
```typescript
public ngOnDestroy() {
  this.subscriptions.unsubscribe();
}
```

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.ts
```diff
@@ -17,18 +17,85 @@
...
+
+  public refreshPayment(paymentId: string) {
+    this.store.dispatch(new PaymentsAction.GetPayment({ organizationId: this.organization.id, paymentId }));
+  }
+
+  private subscribeToActions() {
+    this.getPaymentSubscription = this.actionsSubject.subscribe(data => {
+      if (data.type === PaymentsActionType.GET_PAYMENT_SUCCESS) {
+        // we have a new payment state obtained through manual button, refresh service limits as well
+        this.store.dispatch(new ServiceLimitsAction.GetServiceLimits({ organizationId: this.organization.id }));
```
livthomas: What if user navigates to some other page before `GET_PAYMENT_SUCCESS` action is dispatched? The service limits will not be updated in such case since this subscription will be destroyed together with this component.

Better approach is to have an optional `nextAction?: Action` parameter in the payload of `GetPayment` action and dispatch this `nextAction` (if defined) together with `GetPaymentSuccess` action in the effects class. Then in this component you can simply pass `new ServiceLimitsAction.GetServiceLimits({ organizationId: this.organization.id })` as `nextAction` parameter.

src/app/core/store/organizations/payment/payments.reducer.ts
```diff
@@ -0,0 +1,36 @@
...
+import {PaymentsAction, PaymentsActionType} from "./payments.action";
+import {initialPaymentsState, paymentsAdapter, PaymentsState, selectAllPayments} from "./payments.state";
+
+export function paymentsReducer(state: PaymentsState = initialPaymentsState, action: PaymentsAction.All): PaymentsState {
+  switch (action.type) {
+    case PaymentsActionType.GET_PAYMENT_SUCCESS:
+      return paymentsAdapter.upsertOne({id: action.payload.payment.id, changes: action.payload.payment}, state);
+    case PaymentsActionType.GET_PAYMENTS_SUCCESS:
+      let updates = [];
+      action.payload.payments.forEach(payment => updates.push({ id: payment.id, changes: payment }));
```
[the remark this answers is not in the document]
marvec: Will modify.

src/app/workspace/organization/detail/payments-panel/payments-order/payments-order.component.html
```diff
@@ -20,6 +25,11 @@ <h5 i18n="@@organizations.tab.detail.order.get">Get Lumeer for Business now!</h5
       </slider>
     </div>
   </div>
+  <div class="form-group">
+    <label *ngIf="trial" for="startFrom" i18n="@@organizations.tab.detail.order.start">Subscription Start Date</label>
+    <label *ngIf="!trial" for="startFrom" i18n="@@organizations.tab.detail.order.start">Subscription Extension Start Date</label>
+    <input class="form-control" type="date" [(ngModel)]="startDateText" (input)="updateStartDate($event)" required id="startFrom" name="startDate" />
```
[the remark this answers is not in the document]
marvec: Will keep date input as agreed on the meeting.

src/app/core/store/organizations/service-limits/service-limits.state.ts
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {createEntityAdapter, EntityState} from "@ngrx/entity";
+import {createSelector} from "@ngrx/store";
+import {AppState} from "../../app.state";
+import {ServiceLimitsModel} from "./service-limits.model";
+
+export interface ServiceLimitsState extends EntityState<ServiceLimitsModel> {
```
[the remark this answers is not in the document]
marvec: There will be service limits for each organization. We can also load service limits for future date when they are planned to be valid in the future.

src/app/workspace/organization/detail/contact-form/contact-form.component.html
```diff
@@ -64,7 +64,9 @@ <h5 i18n="@@organizations.tab.detail.contact.invoicingContact">Invoicing Contact
       </div>
     </div>
     <div class="text-right">
-      <button type="submit" class="btn btn-success mb-2" i18n="@@organizations.tab.detail.contact.save" (click)="save()" [disabled]="!invoicingContactForm.form.dirty || !invoicingContactForm.valid">Save</button>
+      <button *ngIf="!savingState && invoicingContactForm.form.dirty" type="submit" class="btn btn-success mb-2" i18n="@@organizations.tab.detail.contact.save" (click)="save()" [disabled]="!invoicingContactForm.form.dirty || !invoicingContactForm.valid">Save</button>
+      <button *ngIf="savingState" type="submit" class="btn btn-success mb-2" i18n="@@organizations.tab.detail.contact.saving" [disabled]="true">Saving</button>
+      <button *ngIf="!savingState && !invoicingContactForm.form.dirty" type="submit" class="btn btn-success mb-2" i18n="@@organizations.tab.detail.contact.saved" [disabled]="true">Saved!</button>
```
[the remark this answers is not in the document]
marvec: Yes, and this was intentional. To show the state of the form. For the user to be sure it is all fine and they did not change it by mistake.

src/app/workspace/organization/detail/contact-form/contact-form.component.ts
```diff
@@ -37,12 +40,28 @@ export class ContactFormComponent implements OnInit {
...
+  private contactSaveSuccessSubscription: Subscription;
+
+  constructor(private organizationSettingsComponent: OrganizationSettingsComponent,
+              private actionsSubject: ActionsSubject) { }
 
-  ngOnInit() {
+  public ngOnInit() {
+    this.contactSaveSuccessSubscription = this.actionsSubject.subscribe(data => {
+      if (data.type === ContactsActionType.SET_CONTACT_SUCCESS) {
+        this.savingState = false;
```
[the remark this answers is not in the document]
marvec: Yes! Ok, need to catch failure as well...

src/app/core/store/organizations/payment/payments.state.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import {createEntityAdapter, EntityState} from "@ngrx/entity";
+import {createSelector} from "@ngrx/store";
+import {AppState} from "../../app.state";
+import {PaymentModel} from "./payment.model";
+import {selectSelectedOrganizationId} from "../organizations.state";
+
+export interface PaymentsState extends EntityState<PaymentModel> {
+  lastCreatedPayment: PaymentModel;
```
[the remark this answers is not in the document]
marvec: Can be kept as is until push notifications are introduced.

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.ts
marvec: Good hint for future!

src/app/core/store/organizations/payment/payments.effects.ts
```diff
@@ -0,0 +1,115 @@
...
+export class PaymentsEffects {
+
+  @Effect()
+  public getPayments$: Observable<Action> = this.actions$.pipe(
+    ofType<PaymentsAction.GetPayments>(PaymentsActionType.GET_PAYMENTS),
+    withLatestFrom(this.store$.select(selectOrganizationsDictionary)),
+    skipWhile(([action, organizationsEntities]) => !organizationsEntities[action.payload.organizationId]),
+    mergeMap(([action, organizationsEntities]) => {
+      const organization = organizationsEntities[action.payload.organizationId];
+      return this.organizationService.getPayments(organization.code).pipe(
```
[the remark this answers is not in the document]
marvec: Will fix!

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.html
marvec: Will fix!

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.ts
marvec: I really cannot do that. First, I would need this nextAction in CreatePayment, not GetPayment. However, I do not have access to action object in the phase of mapping the output of calling http service.  Any workaround?

livthomas: Why `CreatePaymentAction`? In the code above you are clearly reacting to `GetPaymentAction` and not `CreatePaymentAction`.

When it comes to the `action` object not being available after the HTTP service call, just look at other effect classes how we pass this object further down the stream.

src/app/workspace/organization/detail/payments-panel/payments-list/payments-list.component.ts
```diff
@@ -17,18 +17,68 @@
...
+      this.organizationSubscription.unsubscribe();
+    }
+
+    if (this.paymentsSubscription) {
+      this.paymentsSubscription.unsubscribe();
+    }
+  }
+
+  public refreshPayment(paymentId: string) {
+    this.store.dispatch(new PaymentsAction.GetPayment({ organizationId: this.organization.id, paymentId, nextAction: new ServiceLimitsAction.GetServiceLimits({ organizationId: null }) }));
```
livthomas: Why do you pass `organizationId` to `GetPayment` action but do not set it in `GetServiceLimits` action?

marvec: Because it is later set in the action, I copied that pattern from a different effects (collections). Should I set it in here?

livthomas: It was done this way in other effects because there were some actions that depend on entity ID which is not known at the time when the first action is dispatched in a component. For example when you create a new document with a new link instance, you first need to create a document, obtain its ID, and then use it to create a new link instance.

But in this case it does not make sense to do it in such a complicated way.

PR #320, opened 2018-04-26
Fix trailing curly bracket in dialogs

justmesr:


livthomas:
It has already been fixed by @marvec in commit 67477721c8b1613679f5c298cceb7b8ded2667aa.

PR #330, opened 2018-04-27
Show organization service level in workspace selector #LMR-841

kubedo8:


src/app/core/dto/service-limits.ts
```diff
@@ -17,8 +17,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import {ServiceLevelType} from './service-level-type';
+
 export interface ServiceLimits {
-  serviceLevel: string,
+  serviceLevel: ServiceLevelType,
```
livthomas: I'm not really sure if this is going to be mapped correctly since string enum members are not reverse mapped in TypeScript (see [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime)).

I would suggest to keep `string` type in this DTO, use `ServiceLevelType` in the model, and make correct mapping in the converter using our own map (as we do for other string enums).

src/app/workspace/organization/detail/payments-panel/payments-order/payments-order.component.ts
```diff
@@ -89,7 +90,7 @@ export class PaymentsOrderComponent implements OnInit {
       .pipe(filter(serviceLimits => !isNullOrUndefined(serviceLimits)))
       .subscribe(serviceLimits => {
         this.serviceLimits = serviceLimits;
-        if (serviceLimits.serviceLevel == 'FREE') {
+        if (serviceLimits.serviceLevel == ServiceLevelType.FREE) {
```
livthomas: `===`

src/app/workspace/workspace-chooser/resource-chooser/resource-chooser.component.html
```diff
@@ -7,7 +7,7 @@ <h3 class="w-100 d-flex px-3" i18n="@@resource.chooser.header">
 
     <div class="px-3" *ngIf="resources.length == 0 && !canCreateResource"
          i18n="@@resource.chooser.empty">
-      There are no {resourceType, select, organization {organizations} project {projects}} and you are not allowed to
+      There are no {resourceType, select, Organization {Organizations} Project {Projects}} and you are not allowed to
```
livthomas: Isn't it going to look strange capitalized? You just need to keep the term outside the curly brackets in the original form and it does not matter what you put inside.

At the same time, you need to change the translations otherwise it will be shown as before.

PR #335, opened 2018-04-27
 Store document post-its order in view config #LMR-474

justmesr:
- [x] - Prepare post-it order attribute that responds to movement
- [ ] - Store the attribute in view config


src/app/view/perspectives/post-it/post-it-perspective.component.ts
```diff
@@ -175,6 +179,20 @@ export class PostItPerspectiveComponent implements OnInit, OnDestroy {
...
   }
 
+  private createConfigHelper() {
+    this.configHelper = new ConfigHelper(this.store);
+    this.configHelper.initialize();
+  }
+
+  public ngAfterViewInit() {
+    this.createLayout();
+    this.changeDetector.detectChanges()
```
livthomas: Why do we call change detector after the layout is created?

src/app/view/perspectives/post-it/post-it-perspective.component.ts
```diff
@@ -373,32 +360,27 @@ export class PostItPerspectiveComponent implements OnInit, OnDestroy {
...
-
-    if (this.infiniteScroll) {
-      this.infiniteScroll.destroy();
-    }
+    this.unsubscribeIfPossible(this.deletionHelper);
+    this.unsubscribeIfPossible(this.navigationHelper);
+    this.unsubscribeIfPossible(this.configHelper);
+    this.unsubscribeIfPossible(this.infiniteScroll);
+    this.unsubscribeIfPossible(this.collectionsSubscription);
+    this.subscriptions.forEach(subscription => subscription.unsubscribe());
```
livthomas: Maybe it would be worth using just two subscriptions here - one for post-its and one for all other things.

src/app/view/perspectives/post-it/util/config-helper.ts
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Store} from '@ngrx/store';
+import {Subscription} from 'rxjs/Subscription';
+import {AppState} from '../../../../core/store/app.state';
+import {PostItConfigModel} from '../../../../core/store/views/view.model';
+import {selectViewPostItConfig} from '../../../../core/store/views/views.state';
+
+export class ConfigHelper {
```
livthomas: What is the point of this class? It seems that it does not do anything special, just adds more code that would otherwise not be needed since a single line with a subscription could be placed directly in the component.

src/app/workspace/organization/detail/payments-panel/payments-panel.component.ts
```diff
@@ -110,7 +110,7 @@ export class PaymentsPanelComponent implements OnInit, OnDestroy {
   private subscribeToActions() {
     this.paymentCreatedSubscription = this.actionsSubject.subscribe(data => {
       if (data.type === PaymentsActionType.CREATE_PAYMENT_SUCCESS) {
-        console.log("pajoment created");
+        console.log("payment created");
```
livthomas: I think @marvec did this on purpose. :)

marvec: Yes, that was on purpose :-) To differentiate from another similar message I used during development. It does not matter. This is going to be removed anyway...

PR #336, opened 2018-04-27
Unify and fix input field on org., proj. cards #LMR-843

kubedo8:


src/app/shared/users/new-user/new-user.component.html
```diff
@@ -1,7 +1,8 @@
 <div class="row mt-4">
-  <div class="col-6 col-md-5 col-lg-4 d-flex flex-row flex-nowrap">
+  <div class="col-6 col-md-5 col-lg-4 d-flex flex-row flex-nowrap justify-content-center">
     <input-box
       class="flex-grow-2"
+      [style.width]="1+'px'"
```
livthomas: Why do we need to do this calculation every time? What about using `[style.width]="'1px'"`?

PR #345, opened 2018-05-04
remove use attribute id instead of fullName #LMR-789

kubedo8:
Not finished yet...


src/app/core/store/collections/collections.reducer.ts
```diff
@@ -55,28 +58,39 @@ export function collectionsReducer(state: CollectionsState = initialCollectionsS
   }
 }
 
+function onCreateAttributeSuccess(state: CollectionsState, action: CollectionsAction.CreateAttributeSuccess): CollectionsState {
+  let attributes = state.entities[action.payload.collectionId].attributes.slice();
+  attributes.push(action.payload.attribute);
```
livthomas: I am not sure but direct `concat()` instead of `slice()` and `push()` might be faster.

src/app/core/store/collections/collections.reducer.ts
```diff
@@ -55,28 +58,39 @@ export function collectionsReducer(state: CollectionsState = initialCollectionsS
...
     attributes.splice(index, 1, action.payload.attribute);
   } else {
     attributes.push(action.payload.attribute); // TODO preserve order
   }
 
-  if (action.payload.attributeId !== action.payload.attribute.id) {
-    attributes = renameChildAttributes(attributes, action.payload.attributeId, action.payload.attribute.id);
+  if(!isNullOrUndefined(oldAttributeCopy)){
+
+  }
```
livthomas: What should happen in this case? I guess nothing but we need to add "null check" to the condition a few lines below.

src/app/core/store/collections/collections.reducer.ts
```diff
@@ -55,28 +58,39 @@ export function collectionsReducer(state: CollectionsState = initialCollectionsS
...
-  const prefix = oldParentId + '.';
+function renameChildAttributes(attributes: AttributeModel[], oldParentName: string, newParentName: string): AttributeModel[] {
+  const prefix = oldParentName + '.';
   return attributes.map(attribute => {
-    if (attribute.id.startsWith(prefix)) {
-      const [, suffix] = attribute.id.split(oldParentId, 2);
-      return {...attribute, id: newParentId + suffix};
+    if (attribute.name.startsWith(prefix)) {
+      const [, suffix] = attribute.name.split(oldParentName, 2);
+      return {...attribute, id: newParentName + suffix};
```
livthomas: `name: newParentName + suffix`

src/app/core/store/documents/documents.effects.ts
```diff
@@ -287,31 +286,32 @@ export class DocumentsEffects {
...
+  const addedAttributeNames = newDocumentAttributeNames.filter(name => !oldDocumentAttributeNames.includes(name));
+  const removedAttributeNames = oldDocumentAttributeNames.filter(name => !newDocumentAttributeNames.includes(name));
 
-  const attributeIds = attributes.map(attribute => attribute.id);
-  const createdAttributes = addedAttributeIds.filter(id => !attributeIds.includes(id))
-    .map(id => ({id, name: extractAttributeName(id), constraints: [], usageCount: 1}));
+  const attributeNames = attributes.map(attribute => attribute.name);
+  const createdAttributes = addedAttributeNames.filter(name => !attributeNames.includes(name))
+    .map(name => ({id: "?????", name, constraints: [], usageCount: 1}));
+  // TODO I don't have attribute id !!!!!!!!!!!!
```
livthomas: As we discussed, we should always create attributes in a collection first and the add some data to these attributes in a document. This can be done as a series of actions (in case of post-it perspective) so the behavior stays the same for the user and it seems like it is created at the same time.

src/app/core/store/documents/document.model.ts
```diff
@@ -37,3 +37,9 @@ export interface DocumentModel {
   correlationId?: string;
 
 }
+
+export interface DocumentDataModel {
+  attributeId?: string,
+  name?: string,
+  value?: any
```
livthomas: Why do we store it this way? Have not we agreed to use just `attributeId` -> `value` map? I can see that we now do many unnecessary complicated mappings.

src/app/core/store/documents/documents.effects.ts
```diff
@@ -292,53 +278,50 @@ export class DocumentsEffects {
...
-    .map(name => ({name: name, constraints: []}));
-}
-
 function createSyncCollectionAction(collection: CollectionModel,
                                     newDocument: DocumentModel,
                                     oldDocument: DocumentModel): CollectionsAction.UpdateSuccess {
-  const newAttributeNames: string[] = newDocument && newDocument.data ? Object.keys(newDocument.data) : [];
-  const oldAttributeNames: string[] = oldDocument && oldDocument.data ? Object.keys(oldDocument.data) : [];
+  const newAttributeIds: string[] = newDocument && newDocument.data ? newDocument.data.map(d => d.attributeId) : [];
+  const oldAttributeIds: string[] = oldDocument && oldDocument.data ? oldDocument.data.map(d => d.attributeId) : [];
```
livthomas: Do we still need this if all attribute definitions must exist before storing their values?

kubedo8: This is just for increasing/decreasing attributes count in collection

src/app/core/store/documents/document.model.ts
kubedo8: I don't think so that there are complicated mapings. It's needed because when we dispatch action for create/update/patch attributeId can be null. So how can I store unitialized data in map? The answer is simple: it is impossible, because key in map can not be null.

src/app/core/store/collections/collections.action.ts
```diff
@@ -66,6 +66,10 @@ export enum CollectionsActionType {
   CREATE_ATTRIBUTE_SUCCESS = '[Collections] Create Attribute :: Success',
   CREATE_ATTRIBUTE_FAILURE = '[Collections] Create Attribute :: Failure',
 
+  CREATE_ATTRIBUTES = '[Collections] Create Attributes',
```
livthomas: Do we still need actions for creating a single attribute after this new one was added?

src/app/core/store/collections/collections.effects.ts
```diff
@@ -295,6 +297,53 @@ export class CollectionsEffects {
...
   );
 
+  @Effect()
+  public createAttributes$: Observable<Action> = this.actions$.pipe(
+    ofType<CollectionsAction.CreateAttributes>(CollectionsActionType.CREATE_ATTRIBUTES),
+    mergeMap(action => {
+      const attributesDto = action.payload.attributes.map(attr => CollectionConverter.toAttributeDto(attr));
+      const correlationIdMap = action.payload.attributes.reduce((acc, attr) => {
+        acc[attr.name] = attr.correlationId;
+        return acc;
```
livthomas: Why do we need correlation IDs if attributes are still matched based on their names?

PR #355, opened 2018-05-06
Save favourite documents & collection and default workspace to backend #LMR-880

kubedo8:


src/app/core/guards/page-not-found.guard.ts
```diff
@@ -19,30 +19,49 @@
...
-    if (w === 'w' && organizationCode && projectCode) {
-      this.router.navigate(['w', organizationCode, projectCode, 'view', 'search']);
-    } else {
-      this.router.navigate(['workspace']);
-    }
+    return this.getDefaultWorkspace().pipe(
+      map(workspace => {
+        const hasWorkspace = workspace && workspace.organizationCode && workspace.projectCode;
+        if (w === 'w' && organizationCode && projectCode && hasWorkspace) {
+          this.router.navigate(['w', workspace.organizationCode, workspace.projectCode, 'view', 'search']);
```
livthomas: This behavior does not make sense to me. You check if both the default workspace and workspace from the URL is set and then redirect it to the default one.

I think it should make redirects in this order:

1. If workspace is defined in the URL, navigate to it.
2. If not and default workspace is defined, navigate to it.
3. Otherwise, navigate to workspace selection page.

What do you think?

src/app/core/guards/view/view-exist.guard.ts
```diff
@@ -72,13 +72,15 @@ export class ViewExistGuard implements CanActivate {
...
     this.notificationsService.error(message);
+    console.log('view not found');
     return false;
   }
 
   private redirectToCorrectPath(view: ViewModel, workspace: Workspace) {
     const perspective = view.perspective ? view.perspective : Perspective.Search;
     const query = QueryConverter.toString(view.query);
     this.router.navigate(['w', workspace.organizationCode, workspace.projectCode, 'view', {vc: view.code}, perspective,], {queryParams: {query}});
+    console.log('redirect to correct path');
```
livthomas: Forgotten `console.log` calls.

src/app/core/store/collections/collections.effects.ts
```diff
@@ -115,10 +114,10 @@ export class CollectionsEffects {
     withLatestFrom(this.store$.select(selectOrganizationByWorkspace)),
     map(([action, organization]) => {
       if (action.payload.error instanceof HttpErrorResponse && action.payload.error.status == 402) {
-        const title = this.i18n({id: 'serviceLimits.trial', value: 'Free Service'});
+        const title = this.i18n({id: 'serviceLimits.trial', value: 'Trial Service'});
```
livthomas: Are these changes intentional or a result of a bad merge?

src/app/core/store/collections/collections.action.ts
```diff
@@ -219,7 +219,7 @@ export namespace CollectionsAction {
   export class AddFavoriteFailure implements Action {
     public readonly type = CollectionsActionType.ADD_FAVORITE_FAILURE;
 
-    public constructor(public payload: { error: any }) {
+    public constructor(public payload: { collectionId: string, error: any }) {
```
livthomas: Why do we still need these actions? Is it not updated with the collection?

src/app/core/store/documents/documents.action.ts
```diff
@@ -31,13 +31,20 @@ export enum DocumentsActionType {
...
   UPDATE_DATA = '[Documents] Update Data',
   PATCH_DATA = '[Documents] Patch Data',
 
+  ADD_FAVORITE = '[Documents] Add Favorite',
+  ADD_FAVORITE_SUCCESS = '[Documents] Add Favorite :: Success',
+  ADD_FAVORITE_FAILURE = '[Documents] Add Favorite :: Failure',
+
+  REMOVE_FAVORITE = '[Documents] Remove Favorite',
+  REMOVE_FAVORITE_SUCCESS = '[Documents] Remove Favorite :: Success',
+  REMOVE_FAVORITE_FAILURE = '[Documents] Remove Favorite :: Failure',
```
livthomas: Why do we have new actions for these? We should not add a new action for every field in document metadata.

src/app/core/rest/global.service.ts
```diff
@@ -34,6 +35,10 @@ export class GlobalService {
     return this.httpClient.get<User>(`${this.apiPrefix()}/currentUser`);
   }
 
+  public saveDefaultWorkspace(defaultWorkspace: DefaultWorkspace): Observable<any> {
+    return this.httpClient.put(`${this.apiPrefix()}/workspace`, defaultWorkspace);
```
livthomas: Why is it not defined in the user service if it modifies a user?

src/app/core/store/users/users.effects.ts
```diff
@@ -152,6 +154,18 @@ export class UsersEffects {
...
   );
 
+  @Effect()
+  public saveDefaultWorkspace$ = this.actions$.pipe(
+    ofType<UsersAction.SaveDefaultWorkspace>(UsersActionType.SAVE_DEFAULT_WORKSPACE),
+    concatMap(action => {
+      const defaultWorkspaceDto = DefaultWorkspaceConverter.toDto(action.payload.defaultWorkspace);
+      return this.globalService.saveDefaultWorkspace(defaultWorkspaceDto).pipe(
+        concatMap(() => Observable.of()),
+        catchError(() => Observable.of())
```
livthomas: Why do we need `concatMap` here?

src/app/shared/post-it-collections/post-it-collections.module.ts
```diff
@@ -0,0 +1,56 @@
...
+    PostItCollectionComponent,
+    PostItCollectionNameComponent,
+    PostItCollectionAddButtonComponent,
+    PostItCollectionImportButtonComponent,
+  ],
+  exports: [
+    PostItCollectionsComponent
+  ]
+})
+export class PostItCollectionsModule {
```
livthomas: +1 on creating a new module for these components.

src/app/core/dto/default-workspace.ts
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export interface DefaultWorkspace {
+
+  organizationCode?: string;
+  organizationId: string;
+  projectCode?: string;
+  projectId: string;
```
livthomas: Why do we need both codes and IDs?

src/app/core/guards/page-not-found.guard.ts
```diff
@@ -19,30 +19,49 @@
...
+    );
 
-    const message = this.i18n({id: 'page.not.found', value: 'Page not found'});
-    this.notificationService.error(message);
+  }
 
-    return false;
+  private getDefaultWorkspace(): Observable<DefaultWorkspaceModel> {
+    return this.store.select(selectCurrentUser).pipe(
+      map(user => user.defaultWorkspace)
```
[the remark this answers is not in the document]
kubedo8: Ok

src/app/core/guards/page-not-found.guard.ts
kubedo8: Ok, that make sense

src/app/core/guards/view/view-exist.guard.ts
kubedo8: Thanks

src/app/core/store/collections/collections.effects.ts
kubedo8: Bad merge :)

src/app/core/store/collections/collections.action.ts
kubedo8: Of course not. We store information about favorite collections and documents in separate collections, because it's individual for each user.

src/app/core/store/documents/documents.action.ts
kubedo8: It's not metadata. See comment above.

src/app/core/rest/global.service.ts
kubedo8: Because `UserService` has `organizationId` in path and there we don't need it.

src/app/core/store/users/users.effects.ts
kubedo8: Maybe it's useless..

src/app/core/dto/default-workspace.ts
kubedo8: Because we are sending `ids` but receive both. Codes are computed before send, so when organization and project is changed, we send fresh code for navigate to correct url.

src/app/core/store/users/users.effects.ts
kubedo8: When I removed it, I got an error.

src/app/core/store/collections/collections.action.ts
livthomas: Oh, I see. But we return this information from the back-end inside collection/document object, right?

src/app/core/rest/global.service.ts
livthomas: I am thinking if this is the right approach. We already have two REST methods which work with users but cannot be in user service because they do not need `organizationId`. Maybe it would be worth changing user service to have `organizationId` only on those methods which need it. (Just an idea.)

src/app/core/store/collections/collections.action.ts
kubedo8: Yes :)

PR #358, opened 2018-05-10
Store document post-its order in view config #LMR-474

justmesr:


livthomas:
I am closing this since there are too many conflicts which are probably hard to fix so we agreed that @kubedo8 will work on this in a new feature branch and use some of these changes.

PR #361, opened 2018-05-17
Repaired collection attributes tab #LMR-790 #LMR-778 #LMR-777

kubedo8:


src/app/collection/settings/tab/attributes/attributes-pipe.ts
```diff
@@ -17,18 +17,18 @@
...
-
-  constructor(constraint: Constraints, listSuggestion: string) {
-    this.type = constraint.type;
-    this.color = constraint.color;
-    this.name = listSuggestion;
+@Pipe({
+  name: 'attributeFilter'
+})
+@Injectable()
+export class AttributesPipe implements PipeTransform {
```
livthomas: I have added many `Attribute*Pipe` classes in my latest pull-request so this name can be confusing. If the pipe is called `attributeFilter`, I would expect class name `AttributeFilterPipe`.

src/app/collection/settings/tab/attributes/collection-attributes.component.html
```diff
@@ -1,3 +1,86 @@
...
+            <i class="fas fa-plus"></i>
+            <span class="font-weight-bold" i18n="@@collection.tab.attributes.attribute.add">Add attribute</span>
+          </button>
+        </td>
+      </tr>
+      <tr class="d-flex" *ngFor="let attribute of attributes | attributeFilter:searchString">
+        <th scope="row"
+            class="icon-hover col-1 d-flex align-items-center justify-content-center"
+            [class.visible]="isDefaultAttribute(attribute)">
+          <i *ngIf="isDefaultAttribute(attribute); else notDefaultIcon"
```
livthomas: Since this table might be quite big and often redrawn by search, I would recommend to create a pipe for getting the default attribute and use it here. `isDefaultAttribute()` method is called several times in this template.

src/app/collection/settings/tab/attributes/collection-attributes.component.html
```diff
@@ -1,3 +1,86 @@
...
+        <td class="col-3">
+          <button type="button"
+                  class="btn btn-sm btn-success float-right"
+                  (click)="onCreateAttribute()">
+            <i class="fas fa-plus"></i>
+            <span class="font-weight-bold" i18n="@@collection.tab.attributes.attribute.add">Add attribute</span>
+          </button>
+        </td>
+      </tr>
+      <tr class="d-flex" *ngFor="let attribute of attributes | attributeFilter:searchString">
```
livthomas: I think we should use `trackBy` here.

src/app/core/store/collections/collection.util.ts
```diff
@@ -17,10 +17,20 @@
...
 
-import {Attribute} from '../../../../core/dto/attribute';
+import {CollectionModel} from './collection.model';
 
-export interface ConfiguredAttribute extends Attribute {
+export function getDefaultAttributeId(collection: CollectionModel): string {
+  if (collection.defaultAttributeId) {
+    const attributeExist = collection.attributes.findIndex(attr => attr.id === collection.defaultAttributeId) !== -1;
+    if (attributeExist) {
+      return collection.defaultAttributeId;
```
livthomas: What about this?
```typescript
return collection.attributes.find(attr => attr.id === collection.defaultAttributeId);
```

kubedo8: No, because I want to return value only if `defaultAttributeId` is valid.

src/app/collection/settings/tab/attributes/collection-attributes.component.html
kubedo8: Ok

src/app/collection/settings/tab/attributes/collection-attributes.component.html
kubedo8: Good idea.

src/app/collection/settings/tab/attributes/attributes-pipe.ts
kubedo8: Ok

src/app/core/store/collections/collection.util.ts
livthomas: Oh, I see.

src/app/collection/settings/tab/attributes/collection-attributes.component.html
kubedo8: Can you bring me some example how to use pipe in condition?

livthomas: I think the `else` block here is just too complicated. You can have `<i *ngIf="attribute | isDefaultAttribute:collection">` for the first one and `<i *ngIf="!(attribute | isDefaultAttribute:collection)">` for the second one.

livthomas: But maybe your solution has an advantage of executing it just once.

PR #367, opened 2018-05-23
Refactor users page to table #LMR-928

kubedo8:


src/app/shared/users/user-list/user-list.component.ts
```diff
@@ -95,4 +95,13 @@ export class UserListComponent {
     const newPermission = {id: userId, roles: data.roles};
     this.userPermissionChange.emit({newPermission, currentPermission, onlyStore: data.onlyStore});
   }
+
+  public lightenColor(color: string): string {
+    return HtmlModifier.shadeColor(color, .5);
```
livthomas: We use this method in many classes. Maybe we should extract it to a reusable pipe with optional percentage parameter.

src/app/shared/users/user/user.component.html
```diff
@@ -1,42 +1,35 @@
...
-  <div class="col-4 col-md-5 col-lg-6 d-flex flex-row flex-nowrap align-items-center justify-content-center">
-
-    <i *ngFor="let role of getRoles()"
-       class="fas fa-fw {{getIconForRole(role)}} mx-1"
-       [ngClass]="(hasRole(role)) ?  (changeRoles ? 'clickable': 'selected')
+<td class="col-6 d-flex align-items-center">
+
+  <input-box
+    class="flex-grow-2"
+    [style.width]="'1px'"
```
livthomas: In such cases, you can use the following syntax:
```html
[style.width.px]="1"
```

src/app/shared/users/user-list/user-list.component.ts
```diff
@@ -82,7 +79,10 @@ export class UserListComponent {
   }
 
   public getUserPermission(userId: string): PermissionModel {
-    return this.permissions && this.permissions.users && this.permissions.users.find(perm => perm.id === userId);
+    if (isNullOrUndefined(this.resource)) {
+      return;
+    }
+    return this.resource.permissions && this.resource.permissions.users && this.resource.permissions.users.find(perm => perm.id === userId);
```
livthomas: Why not using the shorter syntax?
```typescript
return this.resource && this.resource.persmission && ...
```

src/app/core/store/organizations/organizations.effects.ts
```diff
@@ -189,14 +189,17 @@ export class OrganizationsEffects {
   @Effect()
   public changePermission$ = this.actions$.pipe(
     ofType<OrganizationsAction.ChangePermission>(OrganizationsActionType.CHANGE_PERMISSION),
-    concatMap(action => {
+    withLatestFrom(this.store$.select(selectOrganizationsDictionary)),
+    concatMap(([action, organizations]) => {
+      const organization = organizations[action.payload.organizationId];
```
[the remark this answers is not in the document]
kubedo8: But I don't have action payload on the top level of pipe operators.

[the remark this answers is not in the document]
kubedo8: There is solution but it requires more code and it's less clear: https://stackoverflow.com/questions/49225314/is-there-an-rxjs-operator-similar-to-withlatestfrom-but-with-a-parameter

PR #369, opened 2018-05-26
Add sentry error monitoring

marvec:
Please have a look at package-lock.json. Created with npm@5.6.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7195,8 +7195,8 @@
           "version": "4.11.8",
           "bundled": true,
           "requires": {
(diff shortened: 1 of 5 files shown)


marvec:
But, is the error handler in the core module used for the whole application?

marvec:
The LumeerErrorHandler is currently useless and in the current state of this PR, both error handlers are invoked. I would suggest to remove LumeerErrorHandler at all. I can extract RavenErrorHandler to a separate class if it bothers you much, however, I think we should keep it in the main app module to even handle errors at the top level...

PR #381, opened 2018-06-03
Optimize scripts bundle size #LMR-959

livthomas:
The initial page of the application should now load only around 2.5 MB of JavaScript files.


marvec:
And where us Plotly now?

livthomas:
@marvec It is already contained in the Chart perspective bundle.

PR #383, opened 2018-06-03
Refactor attribute query items in search box #LMR-929

kubedo8:


src/app/shared/directives/editable-div.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+  selector: '[editable-div]',
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => EditableDivDirective),
+      multi: true
+    }
+  ]
+})
+export class EditableDivDirective implements ControlValueAccessor {
```
livthomas: Just a question: What problems does this directive solve?

kubedo8: Reactive forms needs this for two-way-binding.

src/app/shared/search-box/search-box.component.html
```diff
@@ -1,17 +1,23 @@
...
-  <div class="d-flex flex-wrap w-100 pb-1 pr-1">
-    <query-item *ngFor="let queryItem of queryItems; let itemIndex = index"
-                [queryItem]="queryItem"
-                (remove)="onRemoveQueryItem(itemIndex)"
-                class="d-inline-flex mt-1 ml-1 align-items-center">
-    </query-item>
+<div class="form-control w-100 p-0">
+  <form class="d-flex" [formGroup]="form">
+    <div class="d-flex flex-wrap  flex-grow-1 pb-1 pr-1" [style.width.px]="1" formArrayName="queryItems">
+      <query-item *ngFor="let queryItem of queryItems; let itemIndex = index"
```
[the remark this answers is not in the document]
kubedo8: Yes, good catch.

src/app/shared/shared.module.ts
```diff
@@ -83,7 +86,6 @@ import {LinksModule} from './links/links.module';
     DocumentModule,
     PreviewResultsModule,
     LinksModule,
-    PipesModule,
```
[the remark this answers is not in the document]
kubedo8: Because it's duplicated. (see few lines above)

src/app/shared/directives/editable-div.directive.ts
kubedo8: This technique throws deprecation warning in Angular6. See: https://angular.io/api/forms/FormControlName#use-with-ngmodel

I tried to remove [(ngModel)] and directive but it throws error: `no value accessor attached....`. I'm confused now.

livthomas: AFAIK template driven forms use two-way data binding and reactive forms should be manipulated by form controls. But two-way data binding is not a good practice in general.

I think in this case you need to set `control.value` on `queryItem` input change and then react to `(input)` event on the editable `div` and update the query item.

src/app/shared/search-box/query-item/query-item.component.html
```diff
@@ -1,5 +1,5 @@
 <div [style.background-color]="queryItem | queryItemBackground"
-     [style.border-color]="queryItem | queryItemBorder"
+     [style.border-color]="queryItem | queryItemBorder:form.valid"
```
livthomas: I think you need to check if form control is valid and not the whole form. I think all query items will be red now even the valid ones.

kubedo8: It's not the whole form. I'm passing `[form]="queryItemsControl.at(itemIndex)"` into `query-item.component.html` and everything works as expected.

livthomas: Oh, then it is ok. Just the name is confusing.

PR #385, opened 2018-06-08
Collection post it - UI improvements #LMR-963

kubedo8:


src/app/core/store/collections/collections.effects.ts
```diff
@@ -77,6 +77,8 @@ export class CollectionsEffects {
   @Effect()
   public getNames$: Observable<Action> = this.actions$.pipe(
     ofType<CollectionsAction.GetNames>(CollectionsActionType.GET_NAMES),
+    withLatestFrom(this.store$.select(selectCollectionNames)),
+    filter(([action, collectionNames]) => isNullOrUndefined(collectionNames)),
```
livthomas: I am not sure if we want to work with the old data when we check for duplicates on new collection creation. Such performance fixes can be implemented after we have push notifications. Or is it called too many times?

src/app/core/store/navigation/query.util.ts
```diff
@@ -106,3 +107,7 @@ export function conditionFromString(condition: string): ConditionType {
   }
   return null;
 }
+
+export function queryIsNotEmpty(query: QueryModel) {
+  return query && Object.values(query).find(val => val instanceof Array ? val.length > 0 : val);
```
livthomas: I think we should always specify the return parameter of a function unless it is `void`.

src/app/shared/post-it-collections/post-it-collection.component/post-it-collection.component.ts
```diff
@@ -17,31 +17,33 @@
...
 import {Subject, Subscription} from 'rxjs';
 import {isNullOrUndefined} from 'util';
 import {debounceTime, filter} from 'rxjs/operators';
+import {FormControl} from '@angular/forms';
+import {CollectionValidators} from '../../../core/validators/collection.validators';
+import {PostItCollectionNameComponent} from '../collection-name/post-it-collection-name.component';
 
 @Component({
-  selector: 'post-it-collection',
+  selector: '[post-it-collection]',
```
livthomas: What is the reason behind changing this? I think this change just complicates things without any added value.

src/app/shared/search-box/query-item/query-items.converter.ts
```diff
@@ -78,11 +79,15 @@ export class QueryItemsConverter {
     return filters.map(filter => {
       const [collectionId, attributeId, fullCondition] = filter.split(':', 3);
       const collection = this.data.collections.find(collection => collection.id === collectionId);
-      const attribute = collection.attributes.find(attribute => attribute.id === attributeId);
+      const attribute = collection && collection.attributes.find(attribute => attribute.id === attributeId) || null;
```
livthomas: Isn't the result of this `null` by default if nothing is found? Or why is there `|| null` at the end?

src/app/shared/search-box/query-item/query-items.converter.ts
```diff
@@ -102,12 +107,15 @@ export class QueryItemsConverter {
...
     return this.data.linkTypes.filter(linkType => linkTypeIds.includes(linkType.id))
       .map(linkType => {
-        linkType.collections = [
-          this.data.collections.find(collection => collection.id === linkType.collectionIds[0]),
-          this.data.collections.find(collection => collection.id === linkType.collectionIds[1])
-        ];
+        const collection1 = this.data.collections.find(collection => collection.id === linkType.collectionIds[0]);
+        const collection2 = this.data.collections.find(collection => collection.id === linkType.collectionIds[1]);
+        if (!collection1 || !collection2) {
+          return null;
```
livthomas: I would expect some kind of exception here (or at least `console.error()` call) instead of returning `null`. Because now we are just hiding the problems in the application.

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -238,15 +238,35 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
...
-    this.store.dispatch(new CollectionsAction.Create({collection}));
+    this.store.dispatch(new CollectionsAction.Create({collection, callback: this.onCreateCollection()}));
+  }
+
+  private onCreateCollection(): (collection: CollectionModel) => void {
+    const query = this.query;
+    const store = this.store;
+    return collection => {
+      if (queryIsNotEmpty(query)) {
+        store.dispatch(new NavigationAction.AddCollectionToQuery({collectionId: collection.id}));
```
livthomas: I think we should not add collection to the query even if the `collectionIds` array in the query is empty because it means that the user wants to see all collections.

Imagine you have only fulltext in the query, add a new collection and suddenly all other collections will disappear since the query was not empty and you added only the newly created collection to it.

PR #387, opened 2018-06-11
Fix query input #LMR-974

kubedo8:


src/app/shared/search-box/query-item/attribute-condition/attribute-condition.component.ts
```diff
@@ -121,7 +128,22 @@ export class AttributeConditionComponent implements OnInit {
     setTimeout(() => HtmlModifier.setCursorAtTextContentEnd(this.conditionInput.nativeElement));
   }
 
+  public onKeyPress(event: KeyboardEvent) {
+    if (this.isNumber(event.key)) {
```
livthomas: What about `!isNaN(Number(event.key))` ?

PR #388, opened 2018-06-11
RE:Design

marvec:


src/app/core/top-panel/top-panel.component.html
```diff
@@ -1,38 +1,78 @@
...
+          </div>
+        </a>
+      </div>
+      <div class="ml-2 mr-2 d-flex"
+           *ngIf="isWorkspaceSet()">
+
+          <ul class="navbar navbar-triangle my-0 p-0 mr-3 flex-nowrap">
+            <li class="nav-item dropdown">
+              <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown1" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                <i [class]="organization?.icon" [style.color]="organization?.color"></i>&nbsp;{{organization?.code}}
```
livthomas: We should split this into separate components.

src/app/core/top-panel/top-panel.component.ts
```diff
@@ -127,6 +127,20 @@ export class TopPanelComponent implements OnInit {
     return this.isWorkspaceSet() && !this.searchBoxHidden;
   }
 
+  public goToOrganization(page: string) {
+    if (this.workspace && this.workspace.organizationCode) {
+      this.store.dispatch(new RouterAction.Go({path: ['organization', this.workspace.organizationCode, page]}));
```
livthomas: `RouterAction.Go` is defined there only to make it easier to navigate to some other place when you need to do so in the effect classes. It is not necessary to do so in the components. Just more code is executed in this case.

src/app/shared/post-it-collections/post-it-collection.component/post-it-collection.component.scss
```diff
@@ -1,4 +1,13 @@
 @import "../../variables";
+@import '../../bootstrap_var';
```
livthomas: Why do we need to import both these files?

src/app/shared/search-box/search-box.component.html
```diff
@@ -1,3 +1,12 @@
+<!--div class="input-group input-group-lg flex-fill mr-3">
+  <div class="input-group-prepend">
+    <span class="input-group-text" id=""><i class="fas fa-search"></i></span>
+  </div>
+  <div contenteditable="true" class="form-control frame"><span class="badge badge-pill badge-warning">Collection</span></div>
+</div-->
```
livthomas: I think this can be removed unless it is needed.

src/app/shared/users/user/user.component.html
```diff
@@ -11,10 +11,12 @@
 </td>
 
 <td>
+  <!--[ngClass]="(hasRole(role)) ?  (changeRoles ? 'clickable selected': 'selected')
+                               :  (changeRoles ? 'clickable unselected' : 'unselected')"-->
```
livthomas: Can we remove this?

src/app/view/perspectives/post-it/document/post-it-document.component.scss
```diff
@@ -42,13 +55,13 @@
...
-      transition: background-position $animation-color-duration;
+      /*transition: background-position $animation-color-duration;
       background: linear-gradient(to top, #e8e8e8 50%, transparent 50%);
       background-size: 100% 200%;
-      background-position: right top;
+      background-position: right top;*/
 
       &:hover {
-        background-position: right bottom;
+        //background-position: right bottom;
```
livthomas: I think we can remove all these commented styles if they are not needed.

PR #389, opened 2018-06-11
Automatic query submission #LMR-972

kubedo8:


src/app/shared/search-box/search-box.component.ts
```diff
@@ -127,18 +129,42 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
 
   public onAddQueryItem(queryItem: QueryItem) {
     this.queryItems.push(queryItem);
-    this.queryItemsControl.push(queryItemToForm(queryItem));
+
+    const queryItemForm = queryItemToForm(queryItem);
+    this.queryItemsControl.push(queryItemForm);
+
+    if (queryItemForm.valid) {
```
livthomas: If I understand it correctly, `queryItemForm` represents a form group responsible for just a single query item. What happens if you already have an invalid query item in the search box and add a new one which is valid? This condition will be true but the form should not be submitted.

src/app/shared/search-box/search-box.component.ts
```diff
@@ -127,18 +129,42 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
...
+    }
   }
 
   public onRemoveQueryItem(index: number) {
     this.queryItems.splice(index, 1);
+
+    const isQueryItemValid =  this.queryItemsControl.at(index).valid;
     this.queryItemsControl.removeAt(index);
+
+    if(isQueryItemValid) {
```
livthomas: Why do you check the validity of the query item which is being removed? Does it matter if it was valid or not? I think more important information is if the whole search box form is valid at this point.

src/app/shared/search-box/search-box.component.ts
kubedo8: Good point! I will definitely change it :)

src/app/shared/search-box/search-box.component.ts
kubedo8: It is same as above.. thanks for finding that 👍

PR #390, opened 2018-06-11
Show deleted query items in view and search box #LMR-983

kubedo8:


package-lock.json
```diff
@@ -8975,17 +8970,17 @@
...
         "http-auth": "2.4.x",
         "morgan": "^1.6.1",
-        "object-assign": "^4.1.1",
-        "opn": "^5.3.0",
-        "proxy-middleware": "^0.15.0",
-        "send": "^0.16.2",
+        "object-assign": "latest",
+        "opn": "latest",
+        "proxy-middleware": "latest",
+        "send": "latest",
```
livthomas: Why these changes? Do they fix some of the security warnings? If not, we should not include the changes in `package-lock.json` in our commits unless we add a new dependency.

src/app/shared/search-box/query-item/pipes/is-deleted-item.pipe.ts
```diff
@@ -0,0 +1,15 @@
+import {Pipe, PipeTransform} from '@angular/core';
```
livthomas: License header

src/app/shared/search-box/query-item/pipes/query-item-text.pipe.ts
```diff
@@ -0,0 +1,70 @@
+import {Pipe, PipeTransform} from '@angular/core';
```
livthomas: License header

package-lock.json
kubedo8: I just performed `npm-install`... I will revert these changes

PR #395, opened 2018-06-13
Improve search panel alignment #LMR-970

livthomas:
I have improved search panel alignment in order to look good and fill up more space. However, it is just a temporary solution since it throws `ExpressionChangedAfterItHasBeenCheckedError` error and do not react promptly to resizing through `@media` breakpoint which is probably unavoidable in this case.

The biggest problem is how to synchronize the width of the left side of top panel with the right side as well as view controls which are located in totally different component. I have solved it with a new service but it has the drawbacks mentioned above so we will need to figure out better solution in the future.


livthomas:
I have finally solved this by using [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables).

PR #397, opened 2018-06-18
Doc detail (#LMR-989)

marvec:


src/app/shared/document/document-detail/document-detail.component.ts
```diff
@@ -17,379 +17,128 @@
...
-    return this.collection.attributes.find(attr => attr.name === name);
+  public ngOnChanges(changes: SimpleChanges): void {
+    this.renewSubscriptions();
   }
 
-  private patchDocument() {
-    const updatedDocument = this.prepareUpdatedDocument();
+  private renewSubscriptions(): void {
+    this.subscriptions.unsubscribe();
+    this.documentUiService.destroy(this.collection, this.document);
```
kubedo8: Are you sure to reinit `DocumentUi` after each change? What happens to the new data that was generated during the server response?

src/app/core/ui/ui.module.ts
```diff
@@ -0,0 +1,35 @@
...
+@NgModule({
+  imports: [
+    CommonModule,
+    SharedModule
+  ],
+  providers: [
+    DocumentUiService
+  ]
+})
+export class UiModule {
```
livthomas: You don't really need this module. Just define `DocumentUiService` like this:
```typescript
@Injectable({
  providedIn: 'root',
})
```

src/app/shared/preview-results/preview-results.component.ts
```diff
@@ -56,6 +56,7 @@ export class PreviewResultsComponent implements OnInit, OnDestroy {
   public activeIndex = 0;
 
   private allSubscriptions = new Subscription();
+  private dataSubscriptions = new Subscription();
```
livthomas: Where is `dataSubscriptions` except for unsubscribing this empty subscription?

src/app/shared/document/document-detail/document-detail.component.ts
```diff
@@ -17,379 +17,128 @@
...
+import {DocumentsAction} from '../../../core/store/documents/documents.action';
+import {UiRow} from '../../../core/ui/ui-row';
+import DeleteConfirm = DocumentsAction.DeleteConfirm;
 
 @Component({
   selector: 'document-detail',
   templateUrl: './document-detail.component.html',
-  styleUrls: ['./document-detail.component.scss']
+  styleUrls: ['./document-detail.component.scss'],
+  // changeDetection: ChangeDetectionStrategy.OnPush
```
livthomas: Why is this commented? Does not it work with OnPush change detection?

src/app/core/ui/document-ui.ts
```diff
@@ -0,0 +1,470 @@
...
+  private getRowById(id: string): UiRow {
+    if (id) {
+      return this.rows.find(row => row.id === id);
+    }
+
+    return null;
+  }
+
+  private saveChanges(): void {
+    if (this.document) {
```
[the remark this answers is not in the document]
marvec: Interestign, in all programming lectures I ever had, they always advised the way I followed and to omit commands that control flow (return, break, continue, goto...)

src/app/shared/document/document-detail/document-detail.component.ts
marvec: That is a good question, I need to reinvestigate, because the goal was to use it only during change of document or collection.

marvec:
@livthomas What is SPA? "this solution" - can you be more specific? What are the practices for SPAs? What would you to in a simpler way? I am more than happy to omit any unnecessary code, however, I did not find any other way to make it work. Even now it has a drawback - keeping added rows separately as you advised makes the row to jump to a different position when stored on backend.

src/app/core/ui/ui.module.ts
marvec: I thought that you preferred to have fine grained modules... I can easily delete it, I do not need it to be in a module.

src/app/shared/preview-results/preview-results.component.ts
marvec: There is more serious problem to that. Currently, I believe there is a leaking subscription and this is prepared to solve the problem in the subsequent PR.

src/app/shared/document/document-detail/document-detail.component.ts
marvec: Still need to figure out. Will be investigated in the subsequent PR.

src/app/core/ui/ui.module.ts
livthomas: Sure, we want to split application into small modules. But there is no point in creating a module for a single service. Moreover, since Angular 6 it does not need to be declared in any module.

marvec: My idea was that we might end up having LinkUiService etc. as well. Hence the module. Is the module some obstacle or what?

PR #405, opened 2018-06-25
Collection icons (#LMR-995)

marvec:
- Redesigned collection post-it cards
- Updated FontAwesome
- Scrollable icon picker
- Search icons by meta information


src/app/shared/picker/icon-picker/icon-picker.component.html
```diff
@@ -10,25 +10,15 @@
...
-    <a *ngIf="canActivatePage(page + 1)" (click)="selectPage(page + 1)"><i
-      class="clickable fas fa-angle-double-right"></i></a>
-    <i *ngIf="!canActivatePage(page + 1)" class="fas fa-angle-double-right"></i>
+    <i *ngFor="let iconOption of filteredIcons" class="fa-fw {{ iconOption }} text-center rounded icon"
+       [class.selected]="iconOption === selected"
+       [class.active]="iconOption === icon"
+       (mouseenter)="preview(iconOption)"
+       (mouseleave)="preview(null)"
+       (click)="select(iconOption)"
+       id="{{iconId(iconOption)}}"
```
livthomas: This is used inside `ngFor` iterating over a huge array of items so I would highly recommend using a pipe here.

src/app/shared/picker/icon-picker/icon-picker.component.ts
```diff
@@ -143,7 +123,9 @@ export class IconPickerComponent implements OnInit, AfterViewInit {
   }
 
   public filterInput($event: KeyboardEvent): void {
-    this.filter = (<HTMLInputElement>$event.target).value;
+    if ($event) {
+      this.filter = (<HTMLInputElement>$event.target).value;
```
livthomas: This is just a detail but I have seen it a lot in your code. Angular binds event output values to `$event` variable in templates. However, when it comes to TypeScript code, the convention is to name variables in `camelCase` as you would do in Java. And it is recommended by some people to end such names with `$` if it is an observable so you know right away what you can do with it (especially if it is a class variable which you use in a template).

src/app/shared/picker/icon-picker/icons.ts
```diff
@@ -1253,11 +1377,1403 @@ export const brand: string[] = ['fab fa-500px',
...
+  ['toggle-off', ['toggle-off', 'switch']],
+  ['toggle-on', ['toggle-on', 'switch']],
+  ['toolbox', ['toolbox', 'admin', 'container', 'fix', 'repair', 'settings', 'tools']],
+  ['tooth', ['tooth']],
+  ['trade-federation', ['trade-federation']],
+  ['trademark', ['trademark']],
+  ['train', ['train', 'bullet', 'locomotive', 'railway']],
+  ['transgender', ['transgender', 'intersex']],
+  ['transgender-alt', ['transgender-alt']],
+
```
livthomas: This is very inefficient. The time complexity of the solution is `O(N*M)` where `N` is the number of icons and `M` is the maximum number of metadata items per icon. So basically, it executes around `7000` operations on every letter typed to search input.

It can be reduced to `O(1)` by reversing the map so meta words will become keys and the values will contain arrays of icons.

marvec: It is not that easy. You want to search for a substring, so you still need to traverse the whole array of N icons...

PR #410, opened 2018-07-02
 Fix document post-its #LMR-939

kubedo8:


src/app/core/rest/document.service.ts
```diff
@@ -49,6 +49,7 @@ export class DocumentService {
       map(response => response.headers.get('Location').split('/').pop()),
       mergeMap(id => {
         document.id = id;
+        document.creationDate = Date.now();
```
livthomas: This is not related to this PR but we should probably return document instance from back-end on creation.

src/app/core/store/documents/documents.state.ts
```diff
@@ -46,7 +46,7 @@ export const selectDocumentsByQuery = createSelector(selectAllDocuments, selectQ
 );
 
 export const selectDocumentsByCustomQuery = (query: QueryModel) => createSelector(selectAllDocuments,
-  (documents): DocumentModel[] => filterDocuments(sortDocumentsByCreationDate(documents), query)
+  (documents): DocumentModel[] => filterDocuments(sortDocumentsByCreationDate(documents, true), query)
```
livthomas: It would make more sense to have this parametrized so you can pass this order as optional parameter to this selector.

src/app/shared/common.scss
```diff
@@ -225,3 +225,7 @@ ul {
   overflow: hidden;
 }
 
+post-it-document.card:hover .icon {
+  opacity: 1 !important;
```
livthomas: We should place such selectors inside components. Or is there something that prevents us from doing so? Do we want to target all icons inside the child components of `PostItDocumentComponent`? And if so, cannot this be addressed directly in those components?

src/app/view/perspectives/post-it/document/post-it-document-cell/post-it-document-cell.component.html
```diff
@@ -0,0 +1,24 @@
...
+      [title]="model" class="d-flex"
+      (keydown.ArrowLeft)="selectionHelper.moveLeft()"
+      (keydown.ArrowRight)="selectionHelper.moveRight()"
+      (keydown.ArrowUp)="selectionHelper.moveUp()"
+      (keydown.ArrowDown)="selectionHelper.moveDown()"
+      (keydown.F2)="selectionHelper.focusToggle(true)"
+      (keydown.enter)="selectionHelper.focusToggle(true)"
+      (keydown.backspace)="onRemove()"
+      (keydown.delete)="onRemove()"
+      (focus)="onFocus()">
```
livthomas: If you want to reduce the number of DOM elements, you can set all these attributes and event listeners on the component element itself by using `@HostBinding` and `@HostListener` decorators.

src/app/view/perspectives/post-it/document/post-it-document-header/post-it-document-header.component.scss
```diff
@@ -0,0 +1,29 @@
...
+    color: $secondary-hover !important;
+  }
+  &.text-success:hover {
+    color: $success-hover !important;
+  }
+  &.text-danger:hover {
+    color: $danger-hover !important;
+  }
+  &.text-warning:hover {
+    color: $warning-hover !important;
```
livthomas: Why does it not work without all these `!important` things?

src/app/view/perspectives/post-it/document/post-it-document.component.html
```diff
@@ -1,135 +1,65 @@
...
+  [favorite]="getFavorite$() | async"
+  [initialized]="documentModel.id"
+  (remove)="onRemove()"
+  (toggleFavorite)="onToggleFavorite()"
+  (edit)="onEdit()">
+</post-it-document-header>
 
 <div class="card-body p-2">
   <ul class="m-0 p-0" #content>
+    <ng-container *ngFor="let row of getRows$() | async; trackBy:getTrackBy(); let idx = index">
```
livthomas: I am afraid this will generate a new subscription on every change detection cycle. Because `getRows$()` is a function and not class property.

src/app/view/perspectives/post-it/document/post-it-document.component.html
```diff
@@ -1,135 +1,65 @@
...
+      </div>
+    </li>
   </ul>
 </div>
 
-<datalist *ngIf="postItModel"
-          id="{{ suggestionListId() }}">
-  <option *ngFor="let attribute of unusedAttributes()">{{ attribute.name }}</option>
+<datalist *ngIf="documentModel" id="{{ suggestionListId() }}">
+  <option *ngFor="let attribute of getUnusedAttributes$() | async">{{ attribute.name }}</option>
```
livthomas: Can we avoid using these functions in templates? It does not help that it returns an observable.

src/app/core/dto/user.settings.ts
```diff
@@ -20,8 +20,6 @@ import {SizeType} from '../../shared/slider/size-type';
 
 export interface UserSettings {
 
-  defaultOrganization: string;
-  defaultProject: string;
```
livthomas: Why don't we need it anymore?

src/app/view/perspectives/post-it/post-it-perspective.component.html
```diff
@@ -1,35 +1,48 @@
...
+    <div class="layout-item {{getColumnStyle()}} mb-3"
+         layout-item
+         [layout]="layoutManager"
+         *ngFor="let doc of getDocuments(); trackBy:trackByDocument; let idx = index"
+         id="{{perspectiveId}}#{{doc.correlationId || doc.id}}">
+      <post-it-document
+        class="card post-it"
+        [ngClass]="doc.id ? 'initialized': 'uninitialized'"
+        [collection]="getCollection(doc)"
+        [collectionRoles]="getCollectionRoles(doc)"
```
livthomas: I know you have not introduced it here but can we remove these functions from templates?

kubedo8: What do you mean by removing it? How should I pass collection roles to component without this function?

src/app/core/dto/user.settings.ts
kubedo8: Because we're saving it on backend.

src/app/view/perspectives/post-it/document/post-it-document.component.html
kubedo8: What about using `share` operator?

src/app/view/perspectives/post-it/document/post-it-document-header/post-it-document-header.component.scss
kubedo8: I just copied these lines from other component.

src/app/view/perspectives/post-it/post-it-perspective.component.html
livthomas: Using a pipe?

src/app/shared/common.scss
kubedo8: We discussed this with @marvec in live coding two weeks ago. When we tried to place this selector inside component, it doesn't work.

PR #415, opened 2018-07-05
Search tabs improvements #LMR-941

kubedo8:


src/app/shared/post-it-collections/empty-collections/empty-collections.component.html
```diff
@@ -0,0 +1,21 @@
...
+  </div>
+
+  <div *ngIf="project | permissions:'write' | async">
+    <div i18n="@@search.collections.empty.subtitle" class="text-center mb-3">Add your first collection or import some.</div>
+    <post-it-collection-add-button [disabled]="false"
+                                   (clicked)="onNewCollection()">
+    </post-it-collection-add-button>
+
+    <post-it-collection-import-button class="ml-2"
+                                      [disabled]="false" [i18n]="i18n"
```
livthomas: Why do we pass `i18n` here? Cannot it be injected into that component?

src/app/shared/post-it-collections/post-it-collections.component.html
```diff
@@ -1,63 +1,71 @@
...
-
-    <div class="layout-item" layout-item
-         *ngIf="maxShown > 0 && maxShown < collections.length"
-         (click)="onShowAllClicked()"
-         [layout]="layout"
-         [forceLayoutIndex]="maxShown > 0 ? maxShown : null">
-      <div class="see-all d-flex align-items-center justify-content-center cursor-pointer clickable"
-           i18n="@@search.result.more">
-        See all
+<ng-container *ngIf="!isLoading()">
```
livthomas: We should avoid using functions in templates. 🙁

src/app/shared/post-it-collections/post-it-collections.component.ts
```diff
@@ -243,6 +252,10 @@ export class PostItCollectionsComponent implements OnInit, AfterViewInit, OnDest
       }, {});
     });
     this.subscriptions.add(collectionsSubscription);
+
+    const loadedSubscription = this.store.select(selectCollectionsLoaded)
+      .subscribe(loaded => this.collectionsLoaded = loaded);
+    this.subscriptions.add(loadedSubscription);
```
livthomas: Maybe we would need similar mechanism in other components. Then it would make more sense to implement it as another `CollectionsState` property (e.g. `loading:boolean`) which will have its own action to change it. This action will be dispatched with value `true` right before a backend call is made in `CollectionsEffects` and with `false` after `GetSuccess` action.

src/app/shared/warning-message/empty-search/empty-search.component.ts
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Component} from '@angular/core';
+
+@Component({
+  selector: 'empty-search',
+  templateUrl: './empty-search.component.html'
+})
```
livthomas: Can it have `OnPush` change detection?

src/app/shared/warning-message/empty-state/empty-state.component.ts
```diff
@@ -17,14 +17,15 @@
...
 
-import {Document} from '../../../../core/dto/document';
+import {Component, Input} from '@angular/core';
 
-export interface SearchDocument {
+@Component({
+  selector: 'empty-state',
+  templateUrl: './empty-state.component.html'
+})
+export class EmptyStateComponent {
```
livthomas: What about having `OnPush` change detection here as well?

src/app/view/perspectives/search/all/search-all.component.html
```diff
@@ -1,31 +1,39 @@
...
-    Loading...
-  </div>
+<div class="mt-2" *ngIf="dataLoaded$ | async">
 
-  <ng-container *ngIf="!loading()">
-    <div *ngIf="emptySearch()" class="alert alert-warning" i18n="@@search.result.empty">
-      We are sorry, but we have not found anything. Please do a more accurate search.
-    </div>
-    <div *ngIf="collections.length > 0">
+  <ng-container *ngIf="hasData(); else noData">
```
livthomas: Functions... 🙁

src/app/view/perspectives/search/documents/header/search-document-header.component.ts
```diff
@@ -0,0 +1,69 @@
...
+export class SearchDocumentHeaderComponent {
+
+  @Input()
+  public collection: CollectionModel;
+
+  @Input()
+  public document: DocumentModel;
+
+  @Input()
+  public isOpened: Boolean;
```
livthomas: It's not a big deal but we should use primitive types if possible.

src/app/view/perspectives/search/documents/search-document-html-helper.ts
```diff
@@ -0,0 +1,111 @@
...
+    return value as string;
+  }
+}
+
+function searchDocumentGetValuesFromArray(array: any[]): string[] {
+  let values: string[] = [];
+  for (const value of array) {
+    values = values.concat(searchDocumentGetValuesFromAny(value));
+  }
+  return values;
```
livthomas:
```typescript
return array.map(value => searchDocumentGetValuesFromAny(value));
```

src/app/shared/post-it-collections/post-it-collections.component.ts
kubedo8: I just need to know when collections are loaded. The function above has only a bad name..

src/app/view/perspectives/search/documents/header/search-document-header.component.ts
kubedo8: just a typo :)

src/app/shared/post-it-collections/post-it-collections.component.ts
livthomas: Oh, I see. It's something different from what I meant.

PR #430, opened 2018-07-10
Fix url parsing and make it possible to have org. and proj. called view

marvec:


src/index.html
```diff
@@ -23,7 +23,7 @@
     <lmr-app>
       <div style="position: fixed; top: 50%; left: 50%; transform: translate(-50%, -50%); display: flex; flex-direction: column; justify-content: center; align-items: center;">
         <img src="assets/img/lumeer-loading.svg" style="width: 96px; height: 100px;">
-        <h2 style="font-family: sans-serif; color: #2b3e52">Loading...</h2>
+        <h2 style="font-family: Lato, sans-serif; color: #2b3e52">Loading...</h2>
```
livthomas: AFAIK, this only makes sense if you have Lato font installed in your system. Otherwise, the font will not be applied since this code is outside of the main application code and styles (with web fonts) are loaded later with it.

So after this change, if you have the font installed, you will see something completely different than most users.

marvec: Not true, the sans-serif is still there as a second option!

marvec: We definitely need to add Lato font to css on this page as well!!!

PR #431, opened 2018-07-12
 Enable to share views between users #LMR-1005

kubedo8:


src/app/view/view-controls/view-controls.component.ts
```diff
@@ -74,22 +81,40 @@ export class ViewControlsComponent implements OnInit, OnChanges, OnDestroy {
...
     this.name = name;
   }
 
   public ngOnChanges(changes: SimpleChanges): void {
     if (changes.view) {
+      const previousCode = changes.view.previousValue && changes.view.previousValue.code;
+      const currentCode = changes.view.currentValue.code;
+
+      if (previousCode && !currentCode) {
+        this.navigateToUrlWithoutView();
```
livthomas: How can we get to this state where a child component navigates away based on a parameter passed from its parent?

livthomas: My point was that the code above changes the URL based on some state that is passed to it from the parent component while it should be the other way round.

State of the application is usually changed when a user executes some action and if it requires the change of the URL then the only reaction to his action should be navigating to other URL. The  whole application should then react to this change and not vice versa (changing URL after some internal state has changed). URL should be the single source of truth in the application. Otherwise, it will be very error-prone because as you say it is really hard to capture all possible combinations of states.

marvec: Hmmm, we need to change the URL based on the user action. Because the state changes based on what the user does. Except for changes in store, I cannot imagine different state changes. What do you suggest? We need to work on this fast and move forward. What would be the correct approach? Add "selected view" to the store?

PR #436, opened 2018-07-17
Checking permissions in resource menu

kubedo8:


src/app/shared/top-panel/resource-menu/pipes/can-create-resource.pipe.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  public transform(resource: ResourceModel, type: ResourceType): Observable<boolean> {
+    if (type === ResourceType.Organization) {
+      return this.store.select(selectCurrentUser).pipe(
+        map(user => allowedEmails.includes(user.email))
+      );
+    } else if (type === ResourceType.Project) {
+      const project = resource as ProjectModel;
+      return this.store.select(selectOrganizationById(project.organizationId)).pipe(
+        mergeMap(organization => this.permissionsPipe.transform(organization, Role.Write))
```
livthomas: Why cannot we use existing `PermissionPipe` in this case?

src/app/shared/top-panel/resource-menu/resource-detail/resource-status-line/resource-status-line.component.ts
```diff
@@ -50,6 +51,7 @@ export class ResourceStatusLineComponent implements OnChanges {
 
   public ngOnChanges(changes: SimpleChanges): void {
     if (this.organization) {
+      this.store.dispatch(new UsersAction.Get({organizationId: this.organization.id}));
```
livthomas: Is this really the right component which should be responsible for fetching the data? It seems that the organization is passed to it as an input parameter so why don't we fetch the data there?

src/app/shared/top-panel/resource-menu/pipes/can-create-resource.pipe.ts
kubedo8: Because we need to check parent's permissions (in case of organization system permissions).

PR #442, opened 2018-07-22
Show linked documents in detail perspective

kubedo8:


src/app/core/guards/view/view-exist.guard.ts
```diff
@@ -49,7 +49,7 @@ export class ViewExistGuard implements CanActivate {
     }
 
     return this.store.select(selectViewsLoaded).pipe(
-      skipWhile((loaded) => !loaded),
+      filter((loaded) => !loaded),
```
livthomas: The value needs to be negated in order to keep similar behaviour to `skipWhile`.

I have removed this guard in my other PR. Was `skipWhile` operator causing some bug so I should change it in a new guard like you did?

src/app/core/store/navigation/navigation.effects.ts
```diff
@@ -67,7 +67,7 @@ export class NavigationEffects {
     ofType<NavigationAction.RemoveCollectionFromQuery>(NavigationActionType.REMOVE_COLLECTION_TO_QUERY),
     mergeMap(action => this.store$.select(selectQuery).pipe(
       skipWhile(query => !query),
-      first(),
+      take(1),
```
livthomas: Were there any problems with this? Because it should behave the same way.

src/app/shared/document/document-detail/document-detail.component.html
```diff
@@ -34,9 +34,9 @@ <h4>
   </div>
   <div class="col small text-right d-flex align-bottom text-notice">
     <div class="justify-content-end w-100">
-      Version {{document.dataVersion}}<br/>
-      Created on {{document.creationDate | date : 'mediumDate'}} by {{createdBy$ | async}}<br/>
-      Last modified on {{document.updateDate | date : 'mediumDate'}} by {{updatedBy$ | async}}
+      <span>Version {{document.dataVersion}}</span><br/>
+      <span>Created on {{document.creationDate | date : 'mediumDate'}} by {{createdBy$ | async}}</span>
+      <span *ngIf="document.updateDate"><br/>Last modified on {{document.updateDate | date : 'mediumDate'}} by {{updatedBy$ | async}}</span>
```
livthomas: What about adding `i18n` attributes?

src/app/shared/document/document-detail/document-detail.component.ts
```diff
@@ -36,6 +36,9 @@ import {UiRow} from '../../../core/ui/ui-row';
 import DeleteConfirm = DocumentsAction.DeleteConfirm;
 import {Perspective, perspectivesMap} from '../../../view/perspectives/perspective';
 import {PerspectiveService} from '../../../core/perspective.service';
+import {LinkInstancesAction} from '../../../core/store/link-instances/link-instances.action';
+import {selectLinkTypesByDocumentId} from '../../../core/store/link-types/link-types.state';
+import {selectLinkInstancesByTypeAndDocuments} from '../../../core/store/link-instances/link-instances.state';
```
livthomas: I assume these newly added imports are not used anywhere in this component since there are no related changes.

src/app/shared/links/links-list/links-list.component.ts
```diff
@@ -17,21 +17,80 @@
...
-    'name2',
-    'Name3 ja ja lsdjfh aljsdkf aljsd fajsdf ajds flajsdhfla dsfl ajsdf ajsdhf lasd fladsjhf jadsf ajdfhask',
-  ];
+  @Input() public document: DocumentModel;
 
-  public selectedLink: string = 'Name1';
+  @Output() public select = new EventEmitter<{ collection: CollectionModel, document: DocumentModel }>();
+
+  public linkTypes$: Observable<LinkTypeModel[]>;
+  public activeLinkType: LinkTypeModel;
```
livthomas: I cannot select linked document in Detail perspective and I think it is caused by using this property together with `OnPush` change detection. This property should be observable as well so the template will react to the changes properly.

src/app/shared/links/links-list/pipes/link-type-colors.pipe.ts
```diff
@@ -0,0 +1,32 @@
...
+import {Pipe, PipeTransform} from '@angular/core';
+import {LinkTypeModel} from '../../../../core/store/link-types/link-type.model';
+
+@Pipe({
+  name: 'linkTypeColors'
+})
+export class LinkTypeColorsPipe implements PipeTransform {
+
+  public transform(linkType: LinkTypeModel): string[] {
+    return linkType.collections && linkType.collections.map(coll => coll.color) || [];
```
livthomas: Cannot we just pass `linkType.collections` to `ColorsPipe` and fix it so it can handle undefined value?

src/app/shared/links/links-list/pipes/link-type-icons.pipe.ts
```diff
@@ -0,0 +1,32 @@
...
+import {Pipe, PipeTransform} from '@angular/core';
+import {LinkTypeModel} from '../../../../core/store/link-types/link-type.model';
+
+@Pipe({
+  name: 'linkTypeIcons'
+})
+export class LinkTypeIconsPipe implements PipeTransform {
+
+  public transform(linkType: LinkTypeModel): string[] {
+    return linkType.collections && linkType.collections.map(coll => coll.icon) || [];
```
livthomas: The same here. It does not make sense to create new custom pipes for a single use case if we can make use of the existing ones.

src/app/view/perspectives/detail/detail-perspective.component.ts
```diff
@@ -30,20 +32,42 @@ import {QueryModel} from '../../../core/store/navigation/query.model';
...
 
   public selectedDocument: DocumentModel;
 
+  public constructor(private store: Store<AppState>,
+                     private detector: ChangeDetectorRef) {
+  }
+
   public selectCollection(collection: CollectionModel) {
-    this.selectedCollection = collection;
+    this.select(collection, undefined);
```
livthomas: Maybe the second parameter could be optional. 😉

src/app/core/guards/view/view-exist.guard.ts
kubedo8: Maybe in this situation using `skipWhile` and `filter` is the same. But in `effects` using `skipWhile` over `filter` is not a good idea...

src/app/shared/links/links-list/links-list.component.ts
kubedo8: No it works as expected, but you have to do double click to select linked document. In some other PR we will introduce some icon to select document.

src/app/core/store/navigation/navigation.effects.ts
kubedo8: >take(1) emits 1, completes, unsubscribes
>first() emits 1, completes, but doesn't unsubscribe.
according to:
https://stackoverflow.com/questions/42345969/angular-2-using-rxjs-take1-vs-first

livthomas: It seems that description is not very accurate (see the comments there). The difference is better explained [here](https://swalsh.org/blog/rxjs-first-vs-take1/).

PR #487, opened 2018-09-02
 Fix wrong selection and query handling in detail perspective #LMR-1098

kubedo8:


kubedo8:
@livthomas conflicts are now resolved

PR #489, opened 2018-09-03
 Improve table rows rendering performance #LMR-1104

livthomas:


livthomas:
@kubedo8 Look at `IntersectionObserver` and the mechanism of removing elements outside of the viewport from the DOM.

PR #493, opened 2018-09-07
 Add the possibility to add and edit linked documents in detail perspective #LMR-1091

kubedo8:


src/app/shared/input/input-box/input-box.component.ts
```diff
@@ -52,10 +52,12 @@ export class InputBoxComponent implements OnInit {
   @Input() public textAttribute: boolean = false;
   @Input() public innerClass: string = '';
 
-  @Output() public focus: EventEmitter<void> = new EventEmitter();
+  @Output() public focus: EventEmitter<string> = new EventEmitter();
   @Output() public blur: EventEmitter<void> = new EventEmitter();
   @Output() public newValue: EventEmitter<string> = new EventEmitter();
   @Output() public emptyValue: EventEmitter<void> = new EventEmitter();
+  @Output() public enter = new EventEmitter();
+  @Output() public keyDown = new EventEmitter<KeyboardEvent>();
```
livthomas: Why do we need to emit both `keyDown` and `enter` events from this component? Is it not sufficient to emit just `keyDown` and then check if it is Enter or whatever key in the parent component?

src/app/shared/links/links-list/links-list-table/links-list-table.component.ts
```diff
@@ -106,21 +117,69 @@ export class LinksListTableComponent implements OnChanges {
...
 
-  public trackByAttribute(index: number, attribute: AttributeModel): string {
-    return attribute.correlationId || attribute.id;
+  private scrollToNewLinkRow(correlationId: string) {
+    setTimeout(() => {
+      const element = document.getElementById(correlationId);
+      if (element) {
+        const scrollableParent = element.parentElement.parentElement.parentElement;
+        if (scrollableParent) {
+          scrollableParent.scrollTop = Number.MAX_SAFE_INTEGER;
```
livthomas: What about using [Element.scrollIntoView()](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) instead of this complicated structure?

src/app/shared/input/input-box/input-box.component.ts
kubedo8: There's one problem with this approach. I need to catch enter before the input loses focus. But in the `input-box-component.html` `element.blur()` is called on keydown.enter. That's the reason I put separate event before blur is called.

src/app/shared/links/links-list/links-list-table/link-rows-document-ids.pipe.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@Pipe({
+  name: 'linkRowsDocumentIds'
+})
+export class LinkRowsDocumentIdsPipe implements PipeTransform {
+
+  public transform(linkRows: LinkRowModel[]): string[] {
+    return linkRows.reduce((ids, linkRow) => {
+      if (linkRow.document) {
+        ids.push(linkRow.document.id);
```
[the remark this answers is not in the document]
kubedo8: I think my code is faster, isn't it?
But of course, yours is more readable :)

src/app/shared/links/links-list/links-list-table/links-list-table.component.ts
kubedo8: Ok

livthomas:
There is a build error:
```
ERROR in src/app/shared/links/links-list/links-list-table/links-list-table-body/links-list-table-body.component.html(1,5): : Property 'rows' does not exist on type 'LinksListTableBodyComponent'.
```

PR #498, opened 2018-09-12
Add linkTypes tab in collection settings #LMR-1112

kubedo8:
screen shot 2018-09-12 at 15 30 11

@marvec now it looks like on the picture above. Should I add some border on each element to make it prettier?

Delete icon shows on hover...


src/app/collection/settings/tab/link-types/collection-link-types.component.ts
```diff
@@ -17,176 +17,108 @@
...
-    this.linkTypeService.getLinkTypes({collectionIds: [this.collection.id]}).pipe(
-      map(linkTypes => linkTypes.filter(linkType => linkType.collectionIds[0] === this.collection.id))
-    ).subscribe(
-      linkTypes => this.linkTypes = linkTypes.map(linkType => new LinkTypeModel(linkType)),
-      error => this.notificationService.error('Failed fetching LinkTypes')
+  private subscribeData() {
+    this.subscriptions.add(this.store.select(selectCollectionByWorkspace).pipe(
+      filter(collection => !!collection),
+      mergeMap(collection => this.selectLinkTypesForCollection(collection.id)))
+      .subscribe(linkTypes => this.linkTypes = linkTypes)
```
livthomas: What about assigning link types to `linkTypes$` and collections to `collection$` instead of subscribing to their changes? I can see they are both used only in the template.

If you also manage to make `BehaviourSubject` from `searchString$`, you can use `OnPush` change detection in this component since it seems to be well designed.

kubedo8: Good idea :) Just a moment!

livthomas:
Perfect! I hope it works as well. :)

PR #511, opened 2018-09-17
Redirect after organization deletion #LMR-1130

kubedo8:


src/app/shared/resource/header/resource-header.component.html
```diff
@@ -7,10 +7,10 @@
      [style.color]="resource.color" data-toggle="dropdown" (click)="storeIconAndColor($event)"></i>
   <div class="dropdown-menu">
     <color-picker dropdownId="dropdown-header" class="dropdown-item" [(color)]="resource.color"
-                  ></color-picker>
+    ></color-picker>
```
livthomas: I have seen this kind of formatting quite often recently (not only at this project) so we may need to start using [Prettier](https://prettier.io/) soon in order to focus on business logic and avoid such formatting issues.

kubedo8:
@livthomas deletion in ResourcePostItComponent is now fixed

PR #512, opened 2018-09-21
Save document post-it order in view config #LMR-1124

kubedo8:


src/app/core/store/postit/postit.state.ts
```diff
@@ -0,0 +1,36 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {createSelector} from '@ngrx/store';
+import {PostItConfigModel} from '../views/view.model';
+import {AppState} from '../app.state';
+
+export interface PostItState extends PostItConfigModel {
+
+}
```
livthomas: I was thinking that perspectives should have their config stored in `config` field of their state in order to allow storing some runtime configuration outside of this field that does not need to be persisted on view save.

PR #520, opened 2018-09-29
Include view permissions for pos-it perspective #LMR-1157

kubedo8:


src/app/view/perspectives/post-it/document/post-it-document.component.ts
```diff
@@ -48,7 +47,13 @@ export class PostItDocumentComponent implements OnInit, OnDestroy, OnChanges {
   @Output() public remove = new EventEmitter();
   @Output() public sizeChange = new EventEmitter();
 
-  @ViewChild('content') public content: ElementRef;
+  @ViewChild('content') set content(content: ElementRef) {
+    if (content) {
+      this.contentLayout = content;
+      this.disableScrollOnNavigation();
```
livthomas: It seems like this is something that belongs to `ngAfterViewInit` method.

PR #521, opened 2018-09-29
Add methods for processing frontend document meta data #LMR-1175

marvec:


src/app/core/store/documents/documents.action.ts
```diff
@@ -136,6 +139,20 @@ export namespace DocumentsAction {
...
+    public readonly type = DocumentsActionType.UPDATE_META_DATA;
+
+    public constructor(public payload: { document: DocumentModel }) {
+    }
+  }
+
+  export class PatchMetaData implements Action {
+    public readonly type = DocumentsActionType.PATCH_META_DATA;
+
+    public constructor(public payload: { document: DocumentModel }) {
```
livthomas: I am not sure if it the best idea to pass the whole element here since it uses just a subset of its fields and it is not obvious which ones are required.

What about using the following parameters?
```typescript
public payload: { collectionId: string, documentId: string, metaData: {[key: string]: any} }
```

marvec: It seems to me overly complicated and we will always use the single document as a source of all of these parameters. It does not make much sense to me to disassemble before the call. It also introduces extra work every time we must use it. I think that the name of the action tells me which part is needed.

PR #527, opened 2018-10-08
Fix view permissions small bugs #LMR-1193

kubedo8:


src/app/shared/preview-results/preview-results.component.ts
```diff
@@ -143,7 +143,8 @@ export class PreviewResultsComponent implements OnInit, OnDestroy, OnChanges {
     const collectionQuery = {...this.query, collectionIds: [collection.id]};
     this.updateDataSubscription(collectionQuery);
     this.store.dispatch(new DocumentsAction.Get({query: collectionQuery}));
-    this.loaded$ = this.store.select(selectQueryDocumentsLoaded(collectionQuery));
+    this.allSubscriptions.add(this.store.select(selectQueryDocumentsLoaded(collectionQuery))
+      .subscribe(loaded => this.loaded$.next(loaded)));
```
livthomas: Why this change? The previous solution is better since you do not need to handle subscription outside of the template.

kubedo8: For some reason previous observable was not emitting values for the first time when you switch to detail perspective and I have no idea why.

livthomas: In such cases you can always use something like this:
```typescript
.pipe(startWith(false))
```

kubedo8: I will try it in the future, thanks :)

PR #534, opened 2018-10-13
Add a link to the shared view to the view sharing dialog #LMR-1202

kubedo8:


src/app/dialog/dialog.module.ts
```diff
@@ -56,6 +58,8 @@ import {UserRolesPipe} from './share-view/user-roles.pipe';
     DialogBaseComponent,
     ShareViewDialogComponent,
     ShareUserComponent,
+    ShareViewInputComponent,
+    ShareViewCopyComponent,
```
livthomas: What about creating `ShareViewDialogModule`?

src/app/dialog/share-view/share-view-dialog.component.ts
```diff
@@ -65,12 +72,20 @@ export class ShareViewDialogComponent implements OnInit, OnDestroy {
...
 
   public ngOnDestroy() {
     this.subscriptions.unsubscribe();
   }
 
+  public copyToClipboard() {
+    document.addEventListener('copy', this.handler);
+    setTimeout(() => {
+      document.execCommand('copy');
+    });
```
livthomas: What about creating `ClipboardService` with a single `copy()` method we can reuse throughout the app and change the implementation later if we decide to use [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API)?

kubedo8: But current solution use handler which depends on public property from component. So are you sure to extract this code to `ClipboardService` ?

kubedo8: Ok I found solution.

kubedo8:
@livthomas done ✅

PR #548, opened 2018-10-20
 Refactor chart perspective #LMR-1209

kubedo8:


src/app/shared/util.ts
```diff
@@ -0,0 +1,27 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export function hex2rgba(hex: string, opacity: number) {
```
livthomas: There is no need to create a new file for this function since we already work with colors in `html-modifier.ts`. If we decide to move those functions to a different file in the future, we will at least have them at one place.

livthomas: There is no need to create a new file for this function since we already work with colors in `html-modifier.ts`. If we decide to move those functions to a different file in the future, we will at least have them at one place.

src/app/view/perspectives/chart/chart-config/chart-attribute-select/chart-attribute-select.component.ts
```diff
@@ -0,0 +1,68 @@
...
+import {CollectionModel} from '../../../../../core/store/collections/collection.model';
+import {ChartAxisModel, ChartAxisType} from '../../../../../core/store/chart/chart.model';
+
+@Component({
+  selector: 'chart-attribute-select',
+  templateUrl: './chart-attribute-select.component.html',
+  styleUrls: ['../chart-dropdown-select.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ChartAttributeSelectComponent implements OnChanges {
```
livthomas: Is this component still used or has it been replaced by abstract `SelectItemComponent`?

src/app/core/store/chart/chart.model.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface ChartModel {
+  id: string;
+  config?: ChartConfig;
+}
+
+export interface ChartConfig {
+  type: ChartType;
+  xAxis?: ChartAxisModel;
+  y1Axis?: ChartAxisModel;
+  y2Axis?: ChartAxisModel;
```
livthomas: Is this going to be the final number of axes? Wouldn't it be better to store them in an array?

livthomas: Is this going to be the final number of axes? Wouldn't it be better to store them in an array or map?

PR #553, opened 2018-10-29
Include children when filtering documents #LMR-1220

livthomas:
@kubedo8 Please, look at these changes. Since it started to be a little bit complicated, I added unit tests and they immediately discovered a bug in the query converter.


kubedo8:
I think we should create separate selector for this kind of filter, because in other perspectives it doesn't make sense to show children documents (e.g. document post-its).. what do you think?

livthomas:
I have added a new parameter which activates child filtering to the existing selector. Is it ok?

kubedo8:
@livthomas Ok, it's fine

PR #554, opened 2018-10-30
 Make line chart editable #LMR-1210

kubedo8:


src/app/shared/select/select-item/select-item.component.html
```diff
@@ -27,6 +27,7 @@
       <span class="ml-sm-1 text-nowrap">
         {{selectedItem.value}}
       </span>
+      <i *ngIf="removable && !disabled" class="fa-fw fal fa-times mx-1 clear" (click)="$event.preventDefault(); $event.stopPropagation(); onRemove()"></i>
```
livthomas: The code is more readable when we do not call multiple functions directly in the template.

src/app/view/perspectives/chart/chart-visualization/chart-visualization.component.scss
```diff
@@ -0,0 +1,7 @@
+:host ::ng-deep .scatterlayer .trace:last-child .points path {
+  pointer-events: all;
+}
+
+:host ::ng-deep .js-plotly-plot .plotly .modebar {
```
livthomas: I can see that the template of this component contains just a single `<div>` element which is used by Plotly to render the graph. Because of this we can use `ViewEncapsulation.None` for this component in order to avoid using `::ng-deep` here. What do you think?

src/app/view/perspectives/chart/visualizer/plot-maker/plot-maker.ts
```diff
@@ -30,15 +33,65 @@ export abstract class PlotMaker {
 
   protected config: ChartConfig;
 
+  protected onValueChanged?: (valueChange: ValueChange) => void;
+
+  protected onDataChanged?: (dataChange: DataChange) => void;
```
livthomas: What is the difference between value and data change? Is the former change caused by user moving the point and the latter triggered by a new value in the store?

src/app/shared/select/select-item/select-item.component.ts
```diff
@@ -46,11 +46,21 @@ export class SelectItemComponent {
...
   public disabled: boolean;
 
+  @Input()
+  public removable: boolean = false;
+
   @Output()
   public select = new EventEmitter<any>();
 
+  @Output()
+  public removed = new EventEmitter();
```
livthomas: Libraries usually distinguish between `remove` and `removed` events. The former means that the remove action has been triggered while the latter means that the action has successfully completed.

But this event here is about triggering the remove action, right? If so, it would be better to name it `remove` in order to avoid confusion.

src/app/view/perspectives/chart/chart-visualization/chart-visualization.component.scss
kubedo8: That's good idea

src/app/view/perspectives/chart/visualizer/plot-maker/plot-maker.ts
kubedo8: DataChange is triggered while dragging and we update current shown data. DataChange is triggered after drag ends and we update data on the server. But I agree that names are not well choosen.

livthomas: The names are fine. I just wanted to make sure I understand how it works.

kubedo8: Sorry, the second is is ValueChange

kubedo8:
@livthomas done ✅

PR #576, opened 2018-11-21
Push notifications with Pusher #LMR-318

marvec:
Known drawbacks:

* table does not react to changes in links and attributes
* removing access to currently open org. or proj. is not properly handled on frontend


src/app/core/pusher/pusher.service.ts
```diff
@@ -0,0 +1,179 @@
...
+import {ServiceLimitsAction} from '../store/organizations/service-limits/service-limits.action';
+import {ServiceLimitsConverter} from '../store/organizations/service-limits/service-limits.converter';
+import {PaymentsAction} from '../store/organizations/payment/payments.action';
+import {PaymentConverter} from '../store/organizations/payment/payment.converter';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class PusherService implements OnDestroy {
+  private pusher: any;
```
livthomas: As @kubedo8 suggested, better approach would be to add [@types/pusher-js](https://www.npmjs.com/package/@types/pusher-js) as dev dependency to avoid using `any`.

src/app/core/pusher/pusher.service.ts
```diff
@@ -0,0 +1,179 @@
...
+  private subscriptions = new Subscription();
+
+  constructor(private store: Store<AppState>, private authService: AuthService) {
+    if (environment.pusherKey) {
+      this.init();
+    }
+  }
+
+  public init(): void {
+    this.subscriptions.add(
```
livthomas: We do not need to manage this subscription since `take(1)` completes after the first value is emitted. The other reason why this is not needed is that this service is a singleton and so it is destroyed together with the whole application.

src/app/core/pusher/pusher.service.ts
```diff
@@ -0,0 +1,179 @@
...
+        headers: {
+          Authorization: `Bearer ${this.authService.getAccessToken()}`,
+        },
+      },
+    });
+
+    this.channel = this.pusher.subscribe('private-' + user.id);
+
+    this.channel.bind('Organization:create', function(data) {
+      thisRef.store.dispatch(
```
livthomas: You do not need `thisRef` if you use arrow function:
```typescript
this.channel.bind('Organization:create', (data) => {
  this.store.dispatch(
    ...
```

src/app/core/pusher/pusher.service.ts
```diff
@@ -0,0 +1,179 @@
...
+    });
+
+    this.channel.bind('Payment:update', function(data) {
+      thisRef.store.dispatch(
+        new PaymentsAction.GetPaymentSuccess({payment: PaymentConverter.fromDto(data.organizationId, data.entity)})
+      );
+    });
+  }
+
+  public ngOnDestroy(): void {
```
livthomas: What is this method good for? AFAIK it is never called by Angular since this is a service and not a component.

The other thing is that this service is a singleton and so unsubscribing is not really needed. Or does server need to know when a user closes the application so it can free some resources?

src/app/core/pusher/pusher.service.ts
marvec: I tried that but it did not help somehow. It seemed to me like the types were for another library version/deployment

src/app/core/pusher/pusher.service.ts
marvec: Indeed!

src/app/core/pusher/pusher.service.ts
marvec: According to available information, this method should be called on services. I wanted a clean unsubscribe. The messages are otherwise sent my way until I fail to respond to ping.

PR #588, opened 2018-11-27
Add first keyboard shortcuts to table #LMR-1260

livthomas:


src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.ts
```diff
@@ -476,6 +476,23 @@ export class TableDataCellComponent implements OnInit, OnChanges, OnDestroy {
 
   public onKeyDown(event: KeyboardEvent) {
     event[EDITABLE_EVENT] = this.allowedPermissions && this.allowedPermissions.writeWithView;
+
+    if (event.altKey && event.shiftKey) {
+      // TODO check view permissions as well
```
kubedo8: Do we need to check view permissions for actions above?

livthomas: Yes, because some of them change view config and the others change document metadata.

PR #592, opened 2018-11-30
Prepare constraint interfaces #LMR-1281

livthomas:
@marvec @kubedo8 I have prepared interfaces for constraints. Please, check if this is the way we want to work with them.

### DIFF
--- a/src/app/core/model/data/constraint.ts
+++ b/src/app/core/model/data/constraint.ts
@@ -1,0 +1,170 @@
+/*
+ * Lumeer: Modern Data Definition and Processing Platform
+ *
+ * Copyright (C) since 2017 Answer Institute, s.r.o. and/or its affiliates.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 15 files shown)


src/app/core/model/data/constraint.ts
```diff
@@ -0,0 +1,155 @@
...
+  User = 'User',
+}
+
+export interface AddressConstraint {
+  type: ConstraintType.Address;
+  fields: {
+    street: boolean;
+    zip: boolean;
+    city: boolean;
+    state: boolean;
```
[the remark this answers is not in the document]
livthomas: I thought that you would just skip state field (set it to `false`). But you are right, we might need another `region` field in case someone wants to work with this information.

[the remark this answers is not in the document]
livthomas: I am looking at MapQuest Geocoding API we use and it returns `Jihomoravsky` or `Kosicky` in its `state` field. But there is another `county` field we are missing.

src/app/core/model/data/constraint.ts
```diff
@@ -0,0 +1,155 @@
...
+
+import {ColumnFunction} from './column-function';
+
+export enum ConstraintType {
+  Address = 'Address',
+  Boolean = 'Boolean',
+  Coordinates = 'Coordinates',
+  DateTime = 'DateTime',
+  Decimal = 'Decimal',
+  Email = 'Email',
```
kubedo8: What about using `Contact` type which includes `Email`, `Phone` etc.. ?

src/app/core/store/collections/collection.model.ts
```diff
@@ -18,12 +18,14 @@
...
 
 import {ResourceModel} from '../../model/resource.model';
+import {Constraint} from './../../model/data/constraint';
 
 export interface AttributeModel {
   id?: string;
   name: string;
 
-  constraints: string[]; // TODO use complex objects instead
+  constraint?: Constraint; // TODO make compulsory
```
kubedo8: Why array is not used here?

src/app/core/model/data/constraint.ts
livthomas: This PR tries to implement things described in _Use case: Constraints_ document. Feel free to suggest design changes there.

PR #593, opened 2018-11-30
Training videos #LMR-1282

marvec:


src/app/core/store/videos/videos.data.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/* tslint:disable */
+import {environment} from '../../../../environments/environment';
+
+export class VideosData {
+  private static readonly VIDEO_HR = {en: 'etoqX2slVEw', cs: ''};
+  private static readonly VIDEO_GOV = {en: '_msiz33O9Tg', cs: ''};
+
+  private static readonly allVideos = {
+    // 'etoqX2slVEw': 20, // HR
```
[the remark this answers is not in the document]
livthomas: As I am looking at this class more closely, there is no need to have a class here at all since everything is static. And because this is not Java you can define the functions in the file like this:
```typescript
export function getAllVideos() {
  ...
}
```
And constants the way I mentioned in the previous comment. The code would be much cleaner and free of unwanted side-effects.

src/app/app.component.html
```diff
@@ -28,3 +28,4 @@
 
 <notifications></notifications>
 <dialog-base></dialog-base>
+<video-player></video-player>
```
livthomas: I can see that most of the code just duplicates what dialogs do. Why cannot we make use of existing dialogs mechanism? We would then have videos under `(dialog: video/{videoId})` path and we would be able to simply navigate to them by using `DialogService`.

src/app/core/rest/video.service.ts
```diff
@@ -0,0 +1,36 @@
...
+@Injectable()
+export class VideoService {
+  constructor(private httpClient: HttpClient) {}
+
+  public getVideoMetadata(id: string, apiKey: string): Observable<VideoMetaData> {
+    return this.httpClient.get<VideoMetaData>(VideoService.getVideoApi(id, apiKey));
+  }
+
+  private static getVideoApi(id: string, apiKey: string): string {
+    return `https://www.googleapis.com/youtube/v3/videos?part=snippet&id=${id}&key=${apiKey}`;
```
livthomas: We should distinguish between services communicating with our REST API and the ones communicating with external APIs. We started putting the former ones in `core/rest` modules and the latter in `core/api`. I know, the naming is not very clear but let's just keep them separately.

So what I suggest is moving this file to `core/api/video` folder.

src/app/core/store/videos/videos.action.ts
```diff
@@ -0,0 +1,56 @@
...
+  export class LoadVideosFailure implements Action {
+    public readonly type = VideosActionType.LOAD_VIDEOS_FAILURE;
+
+    public constructor(public payload: {error: any}) {}
+  }
+
+  export class RegisterVideos implements Action {
+    public readonly type = VideosActionType.REGISTER_VIDEOS;
+
+    public constructor(public payload: VideoModel[]) {}
```
livthomas: This does not follow action payload structure used everywhere else in the application.

src/app/core/store/videos/videos.data.ts
```diff
@@ -0,0 +1,60 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* tslint:disable */
```
livthomas: What rules are you bypassing here? :slightly_smiling_face:

src/app/core/store/videos/videos.data.ts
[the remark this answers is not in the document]
marvec: Thanks for the tips. Although, the metric of "cleanness" is subjective and there are no side-effects at the moment anyway ;-)

src/app/app.component.html
marvec: I wanted separate component for the dialog window, I did not want to reuse the existing one. And I wanted to keep the possibility for future when we replace the standard dialogs. Is it ok?

src/app/core/store/videos/videos.data.ts
marvec: ' around object keys

src/app/core/store/videos/videos.action.ts
marvec: To be honest, I never noticed any payload structure. Can you tell me the secret please? What would you expect here? `payload: {videos: VideoModel[]}`?

livthomas: Yes, you have already used it in all other actions above.

src/app/app.component.html
livthomas: I simply cannot see any advantage of this approach. If we decide to replace dialog mechanism (as you suggest), we will need to change it on two places since the code will be duplicated. Or do I miss anything?

marvec: And this is exactly it - I don’t think we will be replacing the video player one as it will keep the default look’n’feel. Yes/No dialog is usually a very different from a specialized one... Is it ok?

src/app/dialog/dialog.service.ts
```diff
@@ -24,18 +24,29 @@ import {LinkTypeModel} from '../core/store/link-types/link-type.model';
...
+  providedIn: 'root',
+})
 export class DialogService {
   public callback: any;
   private open: boolean;
 
-  public constructor(private router: Router) {}
+  public constructor(private router: Router, private store: Store<AppState>) {
+    if (environment.videoKey) {
+      this.store.dispatch(new VideosAction.LoadVideos({videos: getAllVideos(), apiKey: environment.videoKey}));
```
livthomas: Doesn't it load videos twice? Once right before the redirect to Auth0 and again after coming back from it?

I don't think it belongs to this constructor or `DialogService` at all. Probably the best solution would be some `VideosGuard` on root view route or elsewhere.

PR #596, opened 2018-12-01
Removed Copy View button #LMR-1265

kubedo8:


src/app/view/view-controls/view-controls.component.ts
```diff
@@ -169,18 +177,21 @@ export class ViewControlsComponent implements OnInit, OnChanges, OnDestroy {
...
       this.navigateToUrlWithoutView({});
     }
   }
 
+  private queryChanged(): boolean {
+    return !areQueriesEqual(this.view.query, this.currentQuery);
+  }
+
+  private configChanged(): boolean {
+    return JSON.stringify(this.currentConfig) !== JSON.stringify(this.view.config[this.currentPerspective]);
```
livthomas: This condition does not work for table perspective. We already have a selector for such comparison which is more sophisticated.

src/app/view/view.component.ts
```diff
@@ -107,29 +115,71 @@ export class ViewComponent implements OnInit, OnDestroy {
...
     this.subscriptions.add(
       combineLatest(this.store.select(selectPerspectiveConfig), this.store.select(selectPerspective))
         .pipe(take(1))
         .subscribe(([config, perspective]) => {
           const view: ViewModel = {...this.view, query: this.query, name, config: {[perspective]: config}, perspective};
 
           if (view.code) {
-            this.updateView(view);
+            if (clone) {
+              this.askToCloneView(view);
```
livthomas: Does the fact that you save a view with only name changed means that you are going to clone it? Because your code seems to work this way.

kubedo8: Yes, look at this https://lumeer.myjetbrains.com/youtrack/issue/LMR-1265#tab=Time%20Tracking

src/app/collection/settings/collection-settings.component.ts
```diff
@@ -143,34 +141,33 @@ export class CollectionSettingsComponent implements OnInit, OnDestroy {
...
-      withLatestFrom(this.store$.pipe(select(selectCollectionByWorkspace))),
-      map(([names, collection]) => (names && names.filter(name => name !== collection.name)) || [])
-    );
+    const sub3 = this.store$
+      .pipe(
+        select(selectPreviousUrl),
+        take(1)
+      )
+      .subscribe(url => (this.previousUrl = url));
+    this.subscriptions.add(sub3);
```
livthomas: I think you don't really need to unsubscribe observables completed by `take` operator.

PR #602, opened 2018-12-07
User notifications #LMR-1292 & #LMR-1293

marvec:


src/app/core/store/user-notifications/user-notification.converter.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  export function toDto(model: UserNotification): UserNotificationDto {
+    return {
+      id: model.id,
+      userId: model.userId,
+      type: UserNotificationType[model.type],
+      read: model.read,
+      data: {}, // cannot be updated by frontend anyway
```
livthomas: Why not making it optional then and not sending it at all?

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.html
```diff
@@ -0,0 +1,200 @@
...
+               class="fa fa-eye-slash text-success ml-1 cursor-pointer"
+               title="Mark as read"
+               title.i18n="@@userNotifications.mark.read"></i>
+          </span>
+        </span>
+        <span class="notification-content read text-notice cursor-pointer"
+              [class.read]="notification.read"
+              [class.text-notice]="notification.read"
+              [class.unread]="!notification.read">
+          <ng-container *ngTemplateOutlet="getTemplate(notification.type); context: { $implicit: notification }"></ng-container>
```
livthomas: This component is way too big. It can be split into at least 7 different components.

For example this very unusual switch can be done in a separate component using Angular's [ngSwitch](https://angular.io/api/common/NgSwitch) directive. And each notification type can have its own component.

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.scss
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    input:checked + .slider {
+      background-color: $success;
+    }
+
+    input:checked + .slider:before {
+      -webkit-transform: translateX(0.95rem);
+      -ms-transform: translateX(0.95rem);
+      transform: translateX(0.95rem);
```
livthomas: You do not have to add this browser-specific styles as they are added automatically by Angular CLI (using [Autoprefixer](https://github.com/postcss/autoprefixer)).

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.html
```diff
@@ -0,0 +1,200 @@
...
+       data-offset="0,10"
+       aria-haspopup="true"
+       aria-expanded="false">
+
+      <span class="fa-stack">
+        <i class="fa fa-bell fa-stack-2x"
+           [class.text-primary]="(unreadNotifications$ | async)?.length > 0"
+           [class.text-secondary]="!(unreadNotifications$ | async) || (unreadNotifications$ | async).length === 0"
+        ></i>
+        <span class="fa-stack-1x" *ngIf="(unreadNotifications$ | async)?.length > 0">
```
livthomas: If you subscribe to the same observable many times in a template, you can use a little trick. Subscribe to it like this in their common ancestor like this:
```html
<div *ngIf="{unread: unreadNotifications$ | async, all: notifications$ | async} as notifications">
  ...
```
... and then use reference them as `notifications.unread` and `notifications.all` where you need them.

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.ts
```diff
@@ -0,0 +1,118 @@
...
+  private unknownTemplate: TemplateRef<any>;
+
+  private organizations$: Observable<Dictionary<OrganizationModel>>;
+
+  private currentWorkspace$: Observable<Workspace>;
+
+  // need to include the notification loader service here for it to initially load notifications and to do that just once
+  constructor(private store: Store<AppState>, private notificationsLoader: UserNotificationsLoaderService) {}
+
+  public ngOnInit(): void {
```
kubedo8: It is not necessary to use `void` return type.

livthomas: I agree with you but I think it is autogenerated like this by the IDE.

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.html
marvec: This was partially laziness, partially saving of time. Maybe some time in the future. I think the component for individual types can be more generic. I suggest to resolve this once we add TaskUserNotification type.

PR #606, opened 2018-12-10
Fix cosmetic changes in videos dropdown, notifications and org/proj selection

marvec:
…hooser


src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.html
```diff
@@ -56,12 +56,12 @@
           </span>
         </span>
         <span class="notification-content read text-notice cursor-pointer"
```
kubedo8: You can remove `read` and `text-notice` class from here, because they are added conditionally.

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.ts
```diff
@@ -106,6 +111,8 @@ export class NotificationsMenuComponent implements OnInit {
   }
 
   private setNotificationRead(notification: UserNotification, read: boolean): void {
+    console.log(notification);
```
kubedo8: Forgotten console.log :)

src/app/shared/top-panel/user-panel/notifications-menu/notifications-menu.component.html
marvec: Good catch!

PR #611, opened 2018-12-14
Add attribute config dialog #LMR-1297

livthomas:


livthomas:
E2E tests fail because they run against backend without changes in my other PR. I ran them locally and they passed.

PR #630, opened 2019-01-05
Fix redirect user notifications #LMR-1361

kubedo8:


src/app/core/service/workspace-select.service.ts
```diff
@@ -0,0 +1,107 @@
...
+          this.checkAndCreateNewProject(organization);
+        }
+      });
+  }
+
+  private goToProject(organization: Organization, project: Project) {
+    if (organization && project) {
+      const nextAction = new RouterAction.Go({path: ['w', organization.code, project.code, 'view', 'search', 'all']});
+      this.store$.dispatch(
+        new ProjectsAction.SwitchWorkspace({organizationId: organization.id, projectId: project.id, nextAction})
```
livthomas: Should not this be done in `WorkspaceGuard` when navigation between projects is detected?

PR #636, opened 2019-01-21
Add constraint data inputs to table #LMR-1332

livthomas:


src/app/shared/data-input/boolean/boolean-data-input.component.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public ngOnChanges(changes: SimpleChanges) {
+    if (changes.indeterminate && !changes.indeterminate.firstChange) {
+      this.setIntermediate();
+    }
+  }
+
+  private setIntermediate() {
+    $(`#${this.inputId}`).prop('indeterminate', this.indeterminate);
```
kubedo8: I recommend you to use `Renderer2`:
```
constructor(private renderer: Renderer2){}

....
this.renderer.setProperty(element, 'indeterminate', this.indeterminate)
....
```

src/app/shared/data-input/data-input.component.html
```diff
@@ -0,0 +1,49 @@
+<ng-container *ngIf="constraint; else textDataInput"
+              [ngSwitch]="constraint.type">
+  <boolean-data-input *ngSwitchCase="'Boolean'"
```
kubedo8: what about defining readonly properties in component and then use it in switch cases?
`public readonly booleanType = ConstraintType.Boolean` etc...

src/app/shared/data-input/data-input.component.html
```diff
@@ -0,0 +1,49 @@
...
+                       [value]="value"
+                       (valueChange)="onValueChange($event)"
+                       (save)="save.emit($event)"
+                       (cancel)="cancel.emit()">
+  </datetime-data-input>
+
+  <number-data-input *ngSwitchCase="'Number'"
+                     [focus]="focus"
+                     [readonly]="readonly"
+                     [constraintConfig]="constraint.config"
```
kubedo8: `contraint?.config` ... also above

src/app/shared/data-input/number/number-data-input.component.html
```diff
@@ -0,0 +1,8 @@
+<input #numberInput
+       [readonly]="readonly"
+       [title]="value"
+       [value]="value"
+       (blur)="onBlur()"
+       (input)="onInput($event)"
+       type="number"
+       class="border-0 w-100 h-100 px-1 bg-transparent">
```
kubedo8: Shouldn't we prevent the user from writing numbers? For example: `pattern="[0-9,.]+"`

src/app/shared/document-hints/document-hints.component.ts
```diff
@@ -94,6 +97,9 @@ export class DocumentHintsComponent implements OnInit, OnChanges {
     if (changes.value) {
       this.filter$.next(this.value);
     }
+    if (changes.collectionId && this.collectionId) {
```
kubedo8: Maybe we can test if collectionId changed.

src/app/shared/pipes/data/datetime-data.pipe.ts
```diff
@@ -0,0 +1,32 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Pipe, PipeTransform} from '@angular/core';
+import * as moment from 'moment';
```
kubedo8: I think this is unused here

src/app/shared/pipes/parse-date.pipe.ts
```diff
@@ -0,0 +1,38 @@
...
+  name: 'parseDate',
+})
+export class ParseDatePipe implements PipeTransform {
+  public transform(value: any): any {
+    if (!value) {
+      return value;
+    }
+
+    const date = new Date(value);
+    if (date.toString() === 'Invalid Date') {
```
kubedo8: moment has function to test if date is valid, but maybe this is also good solution

src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.html
```diff
@@ -8,33 +8,45 @@
...
+<!--(editEnd)="onEditEnd($event)"-->
+<!--(contextmenu)="onContextMenu($event)"-->
+<!--(mousedown)="onMouseDown($event)"-->
+<!--(keydown)="onKeyDown($event)"-->
+<!--[class.uninitialized-column]="!column?.attributeId"-->
+<!--[class.uninitialized-record]="!document?.id && !linkInstance?.id"-->
+<!--[style.width]="column | columnWidth:canManageConfig | pixel"-->
+<!--class="table-border-right table-border-bottom data-input"-->
+<!--data-test="table-data-cell-input">-->
+<!--</div>-->
```
kubedo8: Can not we remove that?

src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.ts
```diff
@@ -87,64 +93,125 @@ export class TableDataCellComponent implements OnInit, OnChanges, OnDestroy {
...
+  public hiddenInputValue$ = new BehaviorSubject<any>('');
 
   private selectedSubscriptions = new Subscription();
+  private subscriptions = new Subscription();
 
   private savingDisabled: boolean;
 
   public constructor(
     private actions$: Actions,
+    private changeDetector: ChangeDetectorRef,
```
kubedo8: Is this even used?

src/app/view/perspectives/table/header/column-group/single-column/table-single-column.component.scss
```diff
@@ -1,6 +1,12 @@
 @import '../../../../../../../styles/variables';
 @import '../../../../../../../styles/bootstrap_var';
 
+:host {
+  & > div {
+    line-height: calc(35px - 0.5rem);
```
kubedo8: Hmmm is it okay to substract `rem` from pixels? `rem` can has various values

src/app/view/perspectives/table/body/rows/cell-group/table-cell-group.component.ts
```diff
@@ -147,4 +147,8 @@ export class TableCellGroupComponent implements OnChanges {
     this.store$.dispatch(new TablesAction.SetCursor({cursor}));
     event.stopPropagation();
   }
+
+  public onAffect() {
+    // just to run change detection
```
kubedo8: I am just curious: Output events run change detection?

livthomas: Yes, of course. We need to run it here because a child component uses `HostBinding` decorator and when its property is changed, the view is not refreshed if you run the change detection only in the child component itself since those properties behave the same as they would if you set them in the parent template.

src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.html
livthomas: Yes, of course. I have kept it here in order to see what are all the actions I need to implement and I forgot to remove it afterwards.

src/app/shared/data-input/number/number-data-input.component.html
livthomas: It is done automatically by browser if you set `type="number"`. Moreover, it also shows you the right keyboard on mobile devices.

src/app/shared/data-input/data-input.component.html
livthomas: There is an `if` condition on the first line of this template so `constraint` cannot be null here.

src/app/shared/data-input/data-input.component.html
livthomas: Thanks, I will change it. But there is even better solution:
```typescript
public readonly constraintType = ConstraintType;
```
...and then we will have something like `*ngSwitchCase="constraintType.Boolean"` in the template.

src/app/dialog/attribute-type/form/constraint-config/number/number-constraint-config-form.component.ts
```diff
@@ -49,6 +49,7 @@ export class NumberConstraintConfigFormComponent implements OnChanges {
   }
 
   private createForm() {
+    this.form.addControl('decimal', new FormControl(this.config && this.config.decimal));
     this.form.addControl('minValue', new FormControl(this.config && this.config.minValue));
     this.form.addControl('maxValue', new FormControl(this.config && this.config.maxValue));
     this.form.setValidators(minMaxValidator('minValue', 'maxValue'));
```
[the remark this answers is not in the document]
livthomas: We cannot really use `FormBuilder` here since it creates a new `FormGroup` and we need to reuse the one provided by the parent component.

src/app/dialog/attribute-type/form/attribute-type-form.component.ts
```diff
@@ -41,7 +41,7 @@ export class AttributeTypeFormComponent implements OnChanges {
   @Output()
   public attributeChange = new EventEmitter<Attribute>();
 
-  public readonly types = Object.keys(ConstraintType).filter(type => ENABLED_CONSTRAINTS.includes(type));
+  public readonly types = ['None'].concat(Object.keys(ConstraintType).filter(type => isConstraintTypeEnabled(type)));
```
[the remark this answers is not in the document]
livthomas: In fact, I had it there before. But in order to stay consistent with the current state when `constraint` is set to `null` in all new attributes, I decided not to have it there. It is only used in a single select and we do not need it in the other parts of the application.

src/app/shared/data-input/boolean/boolean-data-input.component.ts
livthomas: Thanks. I didn't know this one. It works like a charm!

src/app/dialog/attribute-type/form/attribute-type-form.component.ts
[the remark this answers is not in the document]
kubedo8: Okay then

src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.ts
livthomas: It was used before when I tried to run change detection here. But then I realized we need to run it in the parent component to refresh `HostBinding` properties in the view.

src/app/shared/pipes/parse-date.pipe.ts
livthomas: Thanks. I will probably be better to use moment for this since it can parse more variant of ISO date format.

src/app/shared/document-hints/document-hints.component.ts
livthomas: `SimpleChanges` object does not contain `collectionId` if the value is not changed. So this condition is `true` only if there is a real change.

kubedo8: Good point :)

PR #652, opened 2019-02-03
 Implement chart perspective with links & aggregations & sort #LMR-1366

kubedo8:


src/styles/_variables.scss
```diff
@@ -34,7 +34,7 @@ $success: $green !default;
 $info: $cyan !default;
 $warning: $yellow !default;
 $danger: $red !default;
-$light: $gray-200 !default;
+$light: $gray-400 !default;
```
livthomas: This will certainly break some styles in the table where this variable is used.

kubedo8:
@livthomas fixed

src/app/view/perspectives/chart/chart-data/chart-data.component.ts
```diff
@@ -59,38 +89,78 @@ export class ChartDataComponent implements OnChanges {
...
+  public chartData$: Observable<ChartData>;
+  public dataSubject = new Subject<Data>();
 
   constructor(private chartDataConverter: ChartDataConverter) {}
 
-  public ngOnChanges(changes: SimpleChanges) {
+  public ngOnInit() {
+    const closingNotifier = this.dataSubject.pipe(debounceTime(100));
+    this.chartData$ = this.dataSubject.pipe(
+      buffer(closingNotifier),
```
livthomas: It seems like this is just a more complicated alternative to using `debounceTime(100)` directly in this pipe instead of `buffer(closingNotifier)`. Or is there any difference?

kubedo8: I need to know about each change.

kubedo8: Imagine this situation: 
First comes change which indicates that whole chart needs to be updated and then comes another change which indicates about type change. Result is that I need to update whole chart data..

PR #664, opened 2019-02-10
Ask for view copy when config is changed #LMR-1438

kubedo8:


src/app/view/view.component.ts
```diff
@@ -181,10 +181,10 @@ export class ViewComponent implements OnInit, OnDestroy {
     const title = null;
     const message = this.i18n({
       id: 'view.dialog.clone.message',
-      value: 'Do you want to create a copy of the view or just rename?',
+      value: 'Do you want to create a copy of the view or just update existing?',
```
livthomas: I think it should be `update the existing one` in order to sense in English.

kubedo8: Thanks. Resolved

PR #674, opened 2019-02-16
Improve calendar perspective #LMR-1426

kubedo8:


src/app/shared/warning-message/invalid-query/invalid-query.component.scss
```diff
@@ -0,0 +1,3 @@
+:host {
+  width: 100%;
```
livthomas: Styles that affect the surroundings of a component should be defined in the parent component.

If you define some padding in the parent element of this component, this `width` will break styles. Better approach would be to use `w-100` on this component's element in the parent component.

src/app/view/perspectives/calendar/events/header/calendar-header.component.html
```diff
@@ -0,0 +1,57 @@
+<div class="row text-center">
+  <div class="col-lg-4 d-lg-flex align-items-lg-center justify-content-lg-start">
+    <div class="btn-group">
+      <div class="btn btn-primary"
+           mwlCalendarPreviousView
+           [view]="mode"
+           [viewDate]="viewDate"
+           (viewDateChange)="onViewDateChange($event)"
+           i18n="@@previous">
```
livthomas: Maybe it would make sense to provide a more specific translation ID here. We might want to have multiple variants of text `Previous` in the application which might be translated differently, use different case, etc.

kubedo8:
@livthomas resolved

PR #678, opened 2019-02-18
Add attribute function model #LMR-1456

marvec:


src/app/core/store/collections/collection.ts
```diff
@@ -27,13 +27,21 @@ export interface Attribute {
...
   intermediate?: boolean;
 
   correlationId?: string;
 }
 
+export interface AttributeFunction {
+  js?: string;
+  xml?: string;
+  errorReport?: string;
+  timestamp?: number;
```
livthomas: Is this a date and do we want to work with it on frontend? If so, it would make sense to have `Date` type here and convert the value.

marvec: It is rather a technical value for read only on frontend and debugging purposes. Similar as in Blockly rules. There is no point in making it more complicated. We need to see that the value increased over time...

PR #703, opened 2019-03-07
Fix chart range and empty pie chart #LMR-1512

kubedo8:


src/app/view/perspectives/chart/visualizer/plot-maker/axis-draggable-plot-maker.ts
```diff
@@ -43,7 +45,30 @@ export abstract class AxisDraggablePlotMaker extends DraggablePlotMaker {
...
+      this.isAxisCategory(ChartAxisType.Y2)
+    ) {
+      return null;
+    }
+
+    const values = this.chartData.sets.reduce((allValues, set) => {
+      const setValues = set.points
+        .filter(point => isNotNullOrUndefind(point.y) && isNumeric(point.y))
+        .map(point => point.y);
+      return [...allValues, ...setValues];
```
marvec: I once hit a situation where values duplicated in an array created like that. For that purpose I added `uniqueArray` function. Isn't it needed here as well? Just trying to sell my luxurious stuff (the function)...

kubedo8: It's not necessary to use it here :)

kubedo8: It's not necessary to use it here :)

PR #709, opened 2019-03-10
Optimistic updates of user data #LMR-1524

marvec:


src/app/core/store/documents/documents.reducer.ts
```diff
@@ -51,6 +61,33 @@ export function documentsReducer(
...
+      },
+    },
+    state
+  );
+}
+
+function updateDocument(state: DocumentsState, action: DocumentsAction.UpdateData): DocumentsState {
+  const originalDocument = state.entities[action.payload.document.id];
+  if (!action.payload.originalDocument) {
+    action.payload.originalDocument = originalDocument;
```
livthomas: Why do you change the action payload here? Reducer function must not have any side effects!

src/app/core/store/documents/documents.reducer.ts
```diff
@@ -51,6 +61,33 @@ export function documentsReducer(
   }
 }
 
+function patchDocument(state: DocumentsState, action: DocumentsAction.PatchData): DocumentsState {
+  const originalDocument = state.entities[action.payload.document.id];
+  if (!action.payload.originalDocument) {
+    action.payload.originalDocument = originalDocument;
```
livthomas: The same here. This should be done in the effect.

src/app/core/store/documents/documents.action.ts
```diff
@@ -110,19 +110,19 @@ export namespace DocumentsAction {
...
 
-    public constructor(public payload: {error: any}) {}
+    public constructor(public payload: {error: any; originalDocument?: DocumentModel}) {}
   }
 
   export class UpdateData implements Action {
     public readonly type = DocumentsActionType.UPDATE_DATA;
 
-    public constructor(public payload: {document: DocumentModel}) {}
+    public constructor(public payload: {document: DocumentModel; originalDocument?: DocumentModel}) {}
```
livthomas: You do not need to pass original document here. It would be better to get its latest version in the effect pipeline and then put it directly to the failure action.

src/app/core/store/documents/documents.reducer.ts
```diff
@@ -30,8 +30,18 @@ export function documentsReducer(
...
       return addOrUpdateDocument(state, action.payload.document);
+    case DocumentsActionType.UPDATE_DATA_INTERNAL:
+      return updateDocument(state, action);
+    case DocumentsActionType.PATCH_DATA_INTERNAL:
+      return patchDocument(state, action);
     case DocumentsActionType.UPDATE_SUCCESS:
       return addOrUpdateDocument(state, action.payload.document);
+    case DocumentsActionType.UPDATE_FAILURE:
+      if (action.payload.originalDocument) {
+        return documentsAdapter.upsertOne(action.payload.originalDocument, state);
```
livthomas: We should not always update it but instead check the version and proceed only if the document in the store is not newer than the original document.

PR #711, opened 2019-03-11
Percentage constraint #LMR-1525

marvec:


src/app/shared/data-input/percentage/percentage-data-input.component.ts
```diff
@@ -0,0 +1,161 @@
...
+      this.save.emit(this.transformValue(this.percentageInput.nativeElement.value));
+    }
+  }
+
+  private transformValue(value: any): number | string {
+    const text = decimalUserToStore(String(value).trim());
+    if (text.endsWith('%')) {
+      const prefix = text.substring(0, text.length - 1);
+      if (!isNaN(+prefix)) {
+        try {
```
kubedo8: You can extract code to method.

src/app/shared/utils/data.utils.ts
```diff
@@ -87,13 +89,69 @@ export function formatNumberDataValue(value: any, config: NumberConstraintConfig
   return formatUnknownDataValue(value);
 }
 
-export function formatTextDataValue(value: any, config?: TextConstraintConfig): string {
+export function formatPercentageDataValue(value: any, config: PercentageConstraintConfig): string {
+  if ([undefined, null, ''].includes(value)) {
```
kubedo8: Why not `if (!value){....` ?

marvec: When value === 0

PR #719, opened 2019-03-14
Improve number constraint #LMR-1518 #LMR-1519 #LMR-1520

kubedo8:


src/app/shared/utils/data.utils.ts
```diff
@@ -86,7 +87,14 @@ export function formatDateTimeDataValue(value: any, config: DateTimeConstraintCo
...
   // TODO format based on config
-  return formatUnknownDataValue(value);
+  if ([undefined, null, ''].includes(value)) {
+    return '';
+  }
+  const valueBig = convertToBig(value);
+  if (valueBig) {
+    return decimalStoreToUser(valueBig.toFixed());
+  }
+  return String(value);
```
livthomas: All formatting functions should by default use `formatUnknownDataValue` for values they cannot process.

src/app/shared/utils/data.utils.ts
```diff
@@ -171,3 +179,14 @@ export function decimalUserToStore(value: string): string {
 export function decimalStoreToUser(value: string): string {
   return separator === '.' ? value : value.replace('.', separator);
 }
+
+export function convertToBig(value: any): Big {
+  if (isNullOrUndefined(value) || value === '') {
```
livthomas: AFAIK, `isNullOrUndefined` is deprecated. Better condition would be:
```typescript
if (!value && value !== 0) {
```

livthomas: Oh, I can see now that it is our own function.

src/app/shared/data-input/number/number-valid.pipe.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import {Injectable, Pipe, PipeTransform} from '@angular/core';
+import {NumberConstraintConfig} from '../../../core/model/data/constraint';
+import Big from 'big.js';
+import {convertToBig} from '../../utils/data.utils';
+
+@Pipe({
+  name: 'numberValid',
+})
+@Injectable()
```
livthomas: There is no point in using DI for this if it does not use store or any services. It can simply be extracted into a reusable function.

kubedo8:
@livthomas resolved

PR #726, opened 2019-03-18
Add link attributes to table perspective #LMR-1473

livthomas:


src/app/shared/data-input/datetime/datetime-data-input.component.ts
```diff
@@ -93,7 +93,11 @@ export class DatetimeDataInputComponent implements OnChanges {
     if (changes.value && String(this.value).length === 1) {
       // show value entered into hidden input without any changes
       const input = this.dateInput;
-      setTimeout(() => (input.nativeElement.value = this.value));
+      setTimeout(() => {
+        if (input) {
+          input.nativeElement.value = this.value;
```
kubedo8: For the future, there exists shortcut: `(input && (input.nativeElement.value = this.value)))`.

src/app/core/rest/link-type.service.ts
```diff
@@ -17,14 +17,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import {HttpClient, HttpParams} from '@angular/common/http';
+import {HttpClient, HttpParams, HttpResponse} from '@angular/common/http';
```
kubedo8: Maybe unused import?

src/app/shared/data-input/datetime/datetime-data-input.component.ts
livthomas: That's what I wanted to use!

PR #731, opened 2019-03-20
Fix functions

marvec:


src/app/dialog/attribute-function/attribute-function-dialog.component.ts
```diff
@@ -74,7 +74,7 @@ export class AttributeFunctionDialogComponent implements OnInit {
     this.linkType$ = this.selectLinkType();
 
     this.attribute$ = this.selectAttribute(this.selectAttributes(this.collection$, this.linkType$));
-    this.linkTypes$ = this.store$.select(selectAllLinkTypes); // this.selectLinkTypes();
+    this.linkTypes$ = this.selectLinkTypes(this.collection$, this.linkType$); // this.store$.select(selectAllLinkTypes);
```
kubedo8: You can remove unnecessary comment.

PR #740, opened 2019-03-25
Table input on functions, post-it input fix, invalid value underline, top panel #LMR-1457 #LMR-1487

marvec:


src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.html
```diff
@@ -15,6 +15,7 @@
             (save)="onValueSave($event)"
             (cancel)="onCancelEditing()"
             class="d-block h-100"
+            [class.cursor-default]="!isEditable(attribute$ | async)"
```
kubedo8: Use pipe here.

src/app/core/store/collections/collection.util.ts
```diff
@@ -22,7 +22,11 @@ import {Attribute, Collection} from './collection';
 export function isAttributeEditable(attributeId: string, collection: Collection): boolean {
   const attribute =
     attributeId && collection && collection.attributes && collection.attributes.find(attr => attr.id === attributeId);
-  return attribute && (!attribute.function || !attribute.function.js || attribute.function.editable);
+  return isAttributeEditable_(attribute);
+}
+
+export function isAttributeEditable_(attribute: Attribute): boolean {
```
kubedo8: I don't like this suffix :( 
I would suggest you to rename first function to `isCollectionAttributeEditable` and problem will be solved :)

PR #743, opened 2019-03-28
Fix link attr

marvec:


src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.html
```diff
@@ -7,7 +7,7 @@
                       (edit)="onEdit()">
 </table-data-cell-menu>
 
-<data-input [constraint]="constraint$ | async"
+<data-input [constraint]="(attribute$ | async).constraint"
```
kubedo8: I would add `?` symbol after dot just to make sure it won't fail.

marvec: After dot? What it could do? There must be an attribute. If constraint is undefined, the result with ? is the same. Or isn't it?

PR #745, opened 2019-03-30
Color constraint #LMR-1574

marvec:


src/app/shared/data-input/datetime/datetime-data-input.component.ts
```diff
@@ -69,6 +72,8 @@ export class DatetimeDataInputComponent implements OnChanges {
   @ViewChild(BsDatepickerDirective)
   public datePicker: BsDatepickerDirective;
 
+  public locale = moment.locale('cz');
```
kubedo8: Is this used?

src/app/app.component.ts
```diff
@@ -55,7 +63,8 @@ export class AppComponent implements OnInit, AfterViewInit {
     private snotifyService: SnotifyService,
     private store$: Store<AppState>,
     private title: Title,
-    private pusher: PusherService
+    private pusher: PusherService, // to make the service initialize and work
+    public vcRef: ViewContainerRef // for the ngx-color-picker
```
livthomas: Why exactly is this needed if it is not referenced anywhere?

PR #764, opened 2019-04-09
Fix opening datetime picker #LMR-1611

kubedo8:


src/app/shared/date-time/picker/date-time-picker.component.ts
```diff
@@ -115,13 +115,15 @@ export class DateTimePickerComponent implements OnChanges, AfterViewInit, OnDest
...
-      panelClass: ['position-relative', 'w-max-content'],
+      panelClass: ['position-absolute', 'w-max-content'],
+      scrollStrategy: this.overlay.scrollStrategies.reposition(),
       positionStrategy: this.overlay
         .position()
         .flexibleConnectedTo(this.origin)
         .withFlexibleDimensions(false)
         .withViewportMargin(8)
         .withLockedPosition()
+        .withPush(false)
```
livthomas: Maybe the better option would to be add the following positions at the end of the array passed to `withPositions` method:
```typescript
          {
            originX: 'end',
            originY: 'center',
            overlayX: 'start',
            overlayY: 'center',
          },
          {
            originX: 'start',
            originY: 'center',
            overlayX: 'end',
            overlayY: 'center',
          },
```

PR #768, opened 2019-04-10
Complete translations and fix value save on blur #LMR-1644

marvec:


src/app/view/perspectives/table/body/rows/cell-group/data-cell/table-data-cell.component.ts
```diff
@@ -658,7 +666,7 @@ export class TableDataCellComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   public onEdit() {
-    // this.setEditedAttribute();
+    //this.setEditedAttribute();
```
kubedo8: Is this comment needed?

marvec: Tomáš did that, I have no idea why.

PR #773, opened 2019-04-12
Fix document creation when reusing correlation ID #LMR-1651

livthomas:


src/app/core/store/documents/documents.reducer.ts
```diff
@@ -38,6 +38,8 @@ export function documentsReducer(
       return onPatchData(state, action);
     case DocumentsActionType.PATCH_DATA_INTERNAL:
       return patchDocument(state, action);
+    case DocumentsActionType.PATCH_DATA_PENDING:
+      return removeEmptyPendingData(state, action);
```
kubedo8: According to https://github.com/ngrx/platform/issues/162 reducers are executed before effects. Aren't pending data deleted before they are send?

livthomas: I know. As you can see in the `removeEmptyPendingData`, only empty pending data are deleted. And they would not be executed anyway because of the filter in `patchDataPending$` effect.

kubedo8: Ok, great!

PR #779, opened 2019-04-14
Limit free plan resources #LMR-1594

marvec:


src/app/collection/settings/tab/rules/collection-rules.component.ts
```diff
@@ -135,4 +144,21 @@ export class CollectionRulesComponent implements OnInit, OnDestroy {
...
+      .pipe(
+        select(selectOrganizationByWorkspace),
+        map(organization => organization.code),
+        first()
+      )
+      .subscribe(code => {
+        this.store$.dispatch(
+          new RouterAction.Go({
+            path: ['/organization', code, 'detail'],
+            extras: {fragment: 'orderService'},
```
kubedo8: What is fragment?

marvec: Selected tab page, rendered as `#orderService` in url.

kubedo8: Oo I See!

PR #781, opened 2019-04-15
Show empty table row in view #LMR-1657

livthomas:


src/app/core/store/tables/tables.effects.ts
```diff
@@ -600,43 +602,63 @@ export class TablesEffects {
...
-              actions.push(new TablesAction.MoveCursor({direction: Direction.Down}));
-            }
-          }
+          const {collectionId} = table.config.parts[0];
+          return this.store$.pipe(
+            select(selectCollectionById(collectionId)),
+            take(1),
+            mergeMap(collection => this.collectionPermissionsPipe.transform(collection)),
+            take(1),
+            map(permissions => permissions.write || permissions.writeWithView),
```
kubedo8: It's enough to test if user has `writeWithView` permission, but you can leave it as it is.

kubedo8:
Is this approach reactive? What about situation when user lost `write` access to collection during opened table?

src/app/core/store/tables/tables.effects.ts
livthomas: I would leave it here so in the future we can remove the logic for adding empty rows elsewhere.

livthomas:
@kubedo8 You're right. I've just added the code to remove the empty row once you lose the permissions.

kubedo8:
Great!

PR #783, opened 2019-04-15
Disable hints in uninitialized table row #LMR-1654

livthomas:


src/app/shared/document-hints/document-hints.component.html
```diff
@@ -1,26 +1,28 @@
...
-            *ngIf="{value: doc.data[column.attributeId] | dataValue : (collection | collectionAttribute : column.attributeId)?.constraint} as data"
-            [class.font-weight-bold]="column.attributeId === attributeId"
-            [style.width.px]="column.width"
-            [innerHtml]="column.attributeId === attributeId ? (data.value | highlightText: value) : data.value"
-            [title]="doc.data[column.attributeId] || ''"
-            class="text-truncate cursor-default p-1"
-            data-test="document-hint">
-          </div>
-        </ng-container>
+<ng-container *ngIf="linkedDocumentId">
```
kubedo8: Isn't better to add this condition on the whole component from outside?

livthomas: This component is used at various places and that way we would need to add the condition everywhere. Moreover, we could easily forget to add it if we wanted to use the component at some new place.

kubedo8: Ok then

PR #789, opened 2019-04-20
Add support for percentage & datetime constraint in chart #LMR-1609

kubedo8:


src/app/shared/top-panel/search-box/search-box.component.ts
```diff
@@ -90,12 +90,14 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
...
 
   private subscribeToQuery() {
-    const querySubscription = combineLatest(this.store$.pipe(select(selectQuery)), this.loadData(), this.router.events)
+    const querySubscription = combineLatest(this.store$.pipe(select(selectQuery)), this.loadData())
       .pipe(
         debounceTime(100),
-        filter(([query, , event]) => !!query && event instanceof NavigationEnd),
-        tap(([, data]) => (this.queryData = data)),
-        map(([query, data]) => ({queryItems: new QueryItemsConverter(data).fromQuery(query, true), query})),
+        withLatestFrom(this.router.events.pipe(startWith(null))),
```
livthomas: Why are router events needed here?

src/app/view/perspectives/chart/data/convertor/chart-data.ts
```diff
@@ -28,11 +30,12 @@ export interface ChartDataSet {
   id: string;
   points: ChartPoint[];
   color: string;
-  isNumeric: boolean;
   yAxisType: ChartYAxisType;
   name: string;
   draggable: boolean;
   resourceType: ChartAxisResourceType;
+  category: ChartAxisCategory;
+  config?: ConstraintConfig;
```
livthomas: Why is constraint config duplicated here if it can be easily retrieved directly from the store?

src/app/shared/top-panel/search-box/search-box.component.ts
kubedo8: The reason is that `search-box.component` is always visible on the page and sometimes during navigation this observable emits value before query changed (because data are cleaned during navigation) and then it tries to override query which leads to unexpected state. I need to be sure that observable emits data only when `NavigationEnd` event is the latest one.

But let me test this once again just to make sure everything works properly!

src/app/view/perspectives/chart/data/convertor/chart-data.ts
kubedo8: `chart-data-convertor` converts all data to format for chart. I think it's better to do everything in one place. Other reason is that I want chart to be separated from other components and data fetching.

kubedo8:
@livthomas I tested it once again and everything works properly!

PR #791, opened 2019-04-22
Prevent having multiple inputs in edit mode #LMR-1675

livthomas:


src/app/shared/date-time/input/date-time-input.component.ts
```diff
@@ -91,6 +91,10 @@ export class DateTimeInputComponent implements OnChanges {
   }
 
   public onValueChange(date: Date) {
+    // TODO
```
marvec: What does this TODO mean here? What should happen here? And why isn't it finished?

PR #793, opened 2019-04-23
Fixes of table #LMR-1677, view as first-class citize #LMR-1586

marvec:


src/app/view/perspectives/search/views/view-detail/view-detail.component.html
```diff
@@ -1,28 +1,44 @@
...
+            {{view.name}}
+          </span>
 
-          <i *ngIf="view | permissions:viewType:'manage' | async"
-             class="fas fa-fw fa-trash text-danger ml-auto cursor-pointer"
-             (click)="$event.stopPropagation(); onDelete()"
-             title="Delete" i18n-title="@@delete"
-             aria-hidden="true">
-          </i>
+          <div class="flex-grow-1 text-right">
```
kubedo8: Isn't better to have growing text (=span)? Also I suggest you to put `*ngIf` condition here instead of repeating it twice in icons.

src/app/view/perspectives/search/views/view-detail/view-detail.component.ts
```diff
@@ -59,6 +64,13 @@ export class ViewDetailComponent implements OnInit, OnChanges {
 
   public readonly viewType = ResourceType.View;
 
+  public sizeXL = SizeType.XL;
+  public sizeL = SizeType.L;
+  public sizeM = SizeType.M;
+  public sizeS = SizeType.S;
```
kubedo8: Why these reduntant constants? You can define constant like this: `public readonly sizeType = SizeType` and then use it with any value in template.

PR #797, opened 2019-04-26
Fix redirect after delete organization

kubedo8:


kubedo8:
I have tried to simulate this https://lumeer.myjetbrains.com/youtrack/issue/LMR-1664, but it seems like everything works properly.

PR #799, opened 2019-04-27
Create user constraint #LMR-1588

livthomas:


src/app/shared/utils/data.utils.ts
```diff
@@ -141,6 +143,8 @@ export function formatDataValue(value: any, constraint?: Constraint): any {
       return formatColorDataValue(value, constraint.config as ColorConstraintConfig);
     case ConstraintType.Boolean:
       return !!value && value !== '0';
+    case ConstraintType.User:
+    // we need to figure out how to format values dependant on external data
```
kubedo8: I know it's hard to do it now, but this is important to correctly display values in chart and other perspectives. Let's discuss possible solutions.

kubedo8:
Did you try user constraint in query? It's very important to try, because inside query is width of input computed by value.

PR #800, opened 2019-04-28
Update templates handling #LMR-1685

marvec:


src/app/view/perspectives/gantt-chart/util/gantt-chart-util.ts
```diff
@@ -192,7 +192,8 @@ export function isGanttConfigChanged(viewConfig: GanttChartConfig, currentConfig
     return true;
   }
 
-  return ganttConfigCollectionsChanged(viewConfig.collections || {}, currentConfig.collections || {});
+  const isChanged = ganttConfigCollectionsChanged(viewConfig.collections || {}, currentConfig.collections || {});
+  return isChanged;
```
kubedo8: I think you can revert these changes back.

marvec: Correct, this was only for debugging purposes!

PR #806, opened 2019-05-02
Kanban perspective #LMR-1664

kubedo8:


src/app/shared/collapsible-sidebar/collapsible-sidebar.component.scss
```diff
@@ -9,6 +9,10 @@
   overflow: auto;
 }
 
+.toggle {
+  opacity: 0.6;
```
marvec: I suggest to have opacity 1.0 when hover. WDYT?

src/app/view/perspectives/chart/chart-perspective.component.ts
```diff
@@ -146,7 +146,7 @@ export class ChartPerspectiveComponent implements OnInit, OnDestroy {
   private subscribeData() {
     this.documents$ = this.store$.pipe(
       select(selectDocumentsByQuery),
-      distinctUntilChanged((x, y) => JSON.stringify(x) === JSON.stringify(y))
+      distinctUntilChanged((x, y) => deepObjectsEquals(x, y))
     );
```
marvec: You could now use isEqual() from lodash (which also has isDeepEquals).

kubedo8: `deepObjectsEquals` uses `isEqual()` from lodash. You did it recently ;)

src/app/shared/collapsible-sidebar/collapsible-sidebar.component.scss
kubedo8: Ok

src/app/view/perspectives/chart/chart-perspective.component.ts
marvec: I see, it is the other way around!

PR #819, opened 2019-05-14
Remove post-it perspective #LMR-1721

kubedo8:


src/app/view/perspectives/perspective.ts
```diff
@@ -21,7 +21,6 @@ export enum Perspective {
   Detail = 'detail',
   Search = 'search',
   Table = 'table',
-  PostIt = 'postit',
```
livthomas: I think we can keep it here if it still needs to be referenced somewhere. We just need to add it to `FilterPerspectivesPipe` besides `SmartDoc`.

kubedo8: It is only needed in this file, so I think it's better to remove it from `Perspective` enum.

PR #826, opened 2019-05-25
Set Google Analytics events and advanced dimensions reporting #LMR-1740

marvec:
### **DO NOT MERGE YET!!!**

### DIFF
--- a/src/app/core/store/collections/collections.effects.ts
+++ b/src/app/core/store/collections/collections.effects.ts
@@ -31,7 +31,7 @@ import {AppState} from '../app.state';
 import {CommonAction} from '../common/common.action';
 import {DocumentModel} from '../documents/document.model';
 import {DocumentsAction, DocumentsActionType} from '../documents/documents.action';
-import {selectNavigation} from '../navigation/navigation.state';
+import {selectNavigation, selectSearchTab, selectWorkspace} from '../navigation/navigation.state';
 import {NotificationsAction} from '../notifications/notifications.action';
 import {selectOrganizationByWorkspace} from '../organizations/organizations.state';
 import {PermissionType} from '../permissions/permissions';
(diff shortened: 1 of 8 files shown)


src/app/core/store/collections/collections.effects.ts
```diff
@@ -142,6 +144,19 @@ export class CollectionsEffects {
     })
   );
 
+  @Effect()
+  public createSuccess$: Observable<Action> = this.actions$.pipe(
+    ofType<CollectionsAction.CreateSuccess>(CollectionsActionType.CREATE_SUCCESS),
+    withLatestFrom(this.store$.pipe(select(selectCollectionsDictionary))),
+    mergeMap(([, collections]) => {
```
livthomas: The proper way of accomplishing what you want using NgRx would be to replace `@Effect()` by `@Effect({dispatch: false})` and `mergeMap` by `tap` operator without any return value.

We already use it like this in `CommonEffects` for example.

marvec: This is what I was looking for. However, when I tried, it does not work. 
```  @Effect({dispatch: false})
  public createSuccess$: Observable<Action> = this.actions$.pipe(
    ofType<CollectionsAction.CreateSuccess>(CollectionsActionType.CREATE_SUCCESS),
    withLatestFrom(this.store$.pipe(select(selectCollectionsDictionary))),
    tap(([, collections]) => {
```
```Error:(148, 10) TS2322: Type 'Observable<Action | [CreateSuccess, Dictionary<Collection>]>' is not assignable to type 'Observable<Action>'.
  Type 'Action | [CreateSuccess, Dictionary<Collection>]' is not assignable to type 'Action'.
    Property 'type' is missing in type '[CreateSuccess, Dictionary<Collection>]' but required in type 'Action'.```

marvec: Never mind, I figured that out.

PR #827, opened 2019-05-27
Switch table to Angular CDK Virtual Scroll #LMR-1603

livthomas:


src/app/view/perspectives/table/table-perspective.component.ts
```diff
@@ -292,10 +300,20 @@ export class TablePerspectiveComponent implements OnInit, OnChanges, OnDestroy {
...
+
+  public onBodyHorizontalScroll(scrollLeft: number) {
+    // need to check if header was not directly scrolled recently, otherwise infinite scroll loop might occur
+    if (this.tableHeader && !this.hasHeaderScrolledRecently()) {
       this.tableHeader.scroll(-scrollLeft);
     }
   }
+
+  private hasHeaderScrolledRecently(): boolean {
+    return this.lastHeaderScroll && Date.now() - this.lastHeaderScroll < 500;
```
kubedo8: Is 500 ms enought? What if user scrolls header and then (fast) scrolls body?

livthomas:
It should work well now. But further testing would be nice.

PR #829, opened 2019-05-29
Fix rules page when switching project #LMR-1749

marvec:


src/app/view/perspectives/table/body/rows/primary-row/hierarchy-cell/table-hierarchy-cell.component.html
```diff
@@ -1,13 +1,13 @@
 <div [contextMenu]="contextMenuComponent?.contextMenu"
      class="text-nowrap">
-  <i *ngFor="let space of spacesBefore"
-     class="far fa-fw fa-ellipsis-v text-secondary align-middle">
+  <i *ngFor="let space of spacesBefore; let idx = index"
```
kubedo8: What is it needed for?

marvec: Forgot...

PR #832, opened 2019-06-02
Add Smartlook recording #LMR-1756

marvec:


src/app/app.component.ts
```diff
@@ -200,6 +203,20 @@ export class AppComponent implements OnInit, AfterViewInit {
...
     this.lazyLoading$.next(false);
   }
+
+  private initSmartlook() {
+    if (environment.smartlookKey) {
+      this.store$
+        .pipe(
+          select(selectCurrentUser),
+          first(),
+          filter(user => !allowedEmails.includes(user.email))
```
livthomas: If you access the application without being logged in, `user` can be `null`. We need to use `filter` operator here the same way as in `setUpExternalServicesUserContext()` method.

src/app/shared/top-panel/workspace-panel/resource-menu/pipes/can-create-resource.pipe.ts
```diff
@@ -31,7 +31,7 @@ import {Project} from '../../../../../core/store/projects/project';
 import {selectCurrentUser} from '../../../../../core/store/users/users.state';
 import {PermissionsPipe} from '../../../../pipes/permissions/permissions.pipe';
 
-const allowedEmails = ['support@lumeer.io', 'mvecera@lumeer.io', 'kubedo8@gmail.com', 'livoratom@gmail.com'];
+export const allowedEmails = ['support@lumeer.io', 'mvecera@lumeer.io', 'kubedo8@gmail.com', 'livoratom@gmail.com'];
```
livthomas: This should be moved to a new file and renamed to something more descriptive (e.g. `superUserEmails`) as we are giving it a global scope now.

PR #838, opened 2019-06-06
Add selection constraint #LMR-1589

livthomas:


src/app/app.component.ts
```diff
@@ -68,16 +68,23 @@ export class AppComponent implements OnInit, AfterViewInit {
     private snotifyService: SnotifyService,
     private store$: Store<AppState>,
     private title: Title,
-    private pusher: PusherService, // to make the service initialize and work
+    private pusherService: PusherService,
```
marvec: I put the comment there so that it is obvious what the purpose is as it was once deleted by mistake. I do not think it is a good dev. practise to delete each others comments.

kubedo8: But now it's used in code and initialized @marvec

PR #844, opened 2019-06-08
Analytics - returning users tracking #LMR-1756

marvec:
Plus a tiny Blockly update


src/app/auth/auth.service.ts
```diff
@@ -87,7 +88,23 @@ export class AuthService {
...
         this.setSession(authResult);
-        this.store.dispatch(new UsersAction.GetCurrentUser());
+        this.store.dispatch(new UsersAction.GetCurrentUserWithLastLogin());
+        if (environment.analytics) {
+          this.store
+            .pipe(
+              select(selectCurrentUser),
+              filter(user => !!user && !!user.lastLoggedIn),
+              first(),
+              timeout(10000)
```
kubedo8: I think this is useless when you add it after `first()` operator. You should move it before it.

src/app/core/store/users/users.effects.ts
```diff
@@ -83,6 +83,22 @@ export class UsersEffects {
...
   );
 
+  @Effect()
+  public getCurrentUserWithLastLogin$: Observable<Action> = this.actions$.pipe(
+    ofType<UsersAction.GetCurrentUserWithLastLogin>(UsersActionType.GET_CURRENT_USER_WITH_LAST_LOGIN),
+    tap(() => this.store$.dispatch(new UsersAction.SetPending({pending: true}))),
+    mergeMap(() =>
+      this.userService.getCurrentUserWithLastLogin().pipe(
+        map(user => convertUserDtoToModel(user)),
+        mergeMap(user => [new UsersAction.GetCurrentUserSuccess({user}), new UsersAction.SetPending({pending: false})]),
```
kubedo8: We use `flatMap` when we return multiple actions, but maybe `mergeMap` is also suitable..

PR #851, opened 2019-06-10
Add support for external users #LMR-1785

livthomas:


src/app/view/perspectives/table/body/rows/cell-group/collapsed-cell/user/user-collapsed-cell.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  public ngOnChanges(changes: SimpleChanges) {
+    if (changes.values && this.values) {
+      this.validValues = this.extractValidValues(this.values);
+    }
+  }
+
+  private extractValidValues(values: any[]): string[] {
+    const validEmailValues = values.filter(value => isEmailValid(value));
+    return Array.from(new Set(validEmailValues));
```
kubedo8: We have method `uniqueValues` which does exactly the same.

PR #861, opened 2019-06-14
Upgrade to Angular 8 #LMR-1770

livthomas:


src/app/core/store/organizations/payment/payments.effects.ts
```diff
@@ -131,9 +129,7 @@ export class PaymentsEffects {
   public createPayment$: Observable<Action> = this.actions$.pipe(
     ofType<PaymentsAction.CreatePayment>(PaymentsActionType.CREATE_PAYMENT),
     mergeMap(action => {
-      const returnUrl =
-        (action.payload.returnUrl && action.payload.returnUrl) ||
-        (this.location as BrowserPlatformLocation).location.href;
+      const returnUrl = (action.payload.returnUrl && action.payload.returnUrl) || window.location.href;
```
livthomas: @marvec Is this what you have tried to accomplish here?

marvec: Smeg knows ;-) Looks so!

kubedo8: There is redundant `action.payload.returnUrl`

marvec: I see, the first should have been `action.payload` to see if it was there.

livthomas: Oh, I see. But it's always non-null so there is no need to check it. TypeScript would complain about it if you created a new instance of `CreatePayment` class without a payload. We rely on it in all actions.

PR #865, opened 2019-06-17
 Add support to override date config in pivot perspective #LMR-1803

kubedo8:


src/app/shared/select/select-constraint-item/select-constraint-item.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  attributeId: string;
+}
+
+@Component({
+  selector: 'select-constraint-item',
+  templateUrl: './select-constraint-item.component.html',
+  styleUrls: ['./select-constraint-item.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SelectConstraintItemComponent {
```
livthomas: This is a very confusing name for this set of components, pipes and interfaces. If I understand it correctly, they have nothing to do with `SelectConstraint`.

src/app/view/perspectives/pivot/pipe/pivot-header-restricted-attributes.pipe.ts
```diff
@@ -0,0 +1,37 @@
...
+@Pipe({
+  name: 'pivotHeaderRestrictedAttributes',
+})
+export class PivotHeaderRestrictedAttributesPipe implements PipeTransform {
+  public transform(
+    attributes: PivotRowColumnAttribute[],
+    currentAttribute: PivotRowColumnAttribute
+  ): SelectConstraintItemId[] {
+    return (attributes || [])
+      .filter(attribute => !currentAttribute || !deepObjectsEquals(attribute, currentAttribute))
```
livthomas: Is `deepObjectsEquals` really needed here? Can't these objects be compared by their IDs?

src/app/shared/select/select-constraint-item/select-constraint-items-formatter.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  private translateDateReadableFormat(format: string): string {
+    const formatType = Object.entries(DateReadableFormatType).find(([, value]) => value === format);
+    if (formatType && formatType[0] === DateReadableFormatType.Weekly) {
+      return this.i18n(
+        {
+          id: 'select.constraint.items.date.format',
+          value: '{format, select, weekly {[W]W YYYY} other {} }',
```
livthomas: Is this the only format that needs translation? Should not we translate all of them?

kubedo8: Yes this is the only one for now.

src/app/shared/select/select-constraint-item/select-constraint-item.component.ts
kubedo8: Ups, somehow I forgot about `SelectConstraint`. I will change it, but I'm not very good at naming classes, interfaces etc :D

src/app/view/perspectives/pivot/pipe/pivot-header-restricted-attributes.pipe.ts
kubedo8: Yes because it's `PivotAttribute` which is totally different from classic `Attribute`. Maybe if I rename variables it will be more clear..

PR #866, opened 2019-06-18
Fix table header scrolling #LMR-1742

livthomas:


src/app/view/perspectives/table/header/column-group/single-column/input/table-column-input.component.ts
```diff
@@ -0,0 +1,122 @@
...
+        event.preventDefault();
+        event.stopPropagation();
+        return;
+      case KeyCode.ArrowUp:
+      case KeyCode.ArrowDown:
+        event.preventDefault();
+        return;
+    }
+
+    if (DISABLED_CHARACTERS.includes(event.key)) {
```
kubedo8: What about case when user paste text with disabled characters here? Is this covered?

src/app/view/perspectives/table/header/table-header.component.ts
```diff
@@ -57,7 +57,11 @@ export class TableHeaderComponent implements OnInit, OnChanges {
   public singleCollection$: Observable<boolean>;
   public cursor: TableHeaderCursor;
 
-  public constructor(private renderer: Renderer2, private store$: Store<AppState>) {}
+  public constructor(
+    private element: ElementRef<HTMLElement>,
+    private renderer: Renderer2,
```
kubedo8: Are these properties needed?

src/app/view/perspectives/table/header/column-group/single-column/input/table-column-input.component.ts
livthomas: Good point!

src/app/view/perspectives/table/header/table-header.component.ts
livthomas: Not anymore.

PR #872, opened 2019-06-19
Fix redirect after project creation #LMR-1804

kubedo8:


src/app/dialog/dialog.service.ts
```diff
@@ -48,9 +48,16 @@ export class DialogService {
     }
   }
 
+  public closeDialogWithoutRedirect() {
+    this.open = false;
```
livthomas: What about cleaning the callback? Would not it be better to add optional `skipRedirect` parameter to `closeDialog` method so we don't need to keep both these methods in sync when something is added to one of them?

src/app/core/store/projects/projects.effects.ts
```diff
@@ -128,6 +128,10 @@ export class ProjectsEffects {
         mergeMap(newProject => {
           const actions: Action[] = [new ProjectsAction.CreateSuccess({project: newProject})];
 
+          if (onSuccess) {
+            actions.push(new CommonAction.ExecuteCallback({callback: () => onSuccess(newProject)}));
```
livthomas: There is a much easier way how to do this now. Have a look at the usages of `createCallbackActions` function.

PR #874, opened 2019-06-20
Add support to format date config in kanban perspective #LMR-1826

kubedo8:


src/app/view/perspectives/kanban/util/kanban-converter.ts
```diff
@@ -0,0 +1,262 @@
...
+
+  private formatKanbanColumnValue(
+    value: any,
+    constraint: Constraint,
+    constraintData: ConstraintData,
+    kanbanAttribute: KanbanAttribute
+  ): any {
+    if (constraint) {
+      if (constraint.type === ConstraintType.User) {
+        return value;
```
livthomas: Is this shown as column title? Don't we want to display user name if it is available? What about using `formatUserDataValue` function?

kubedo8: Yes we do. Look at `kanban-column-header.component`. In order to show user with avatar we need to format it directly in template.

PR #884, opened 2019-06-26
Fix dialog opening mechanism #LMR-1859

livthomas:


src/app/dialog/dialog.service.ts
```diff
@@ -40,34 +43,33 @@ import {Store} from '@ngrx/store';
...
 
-  public constructor(private router: Router, private store: Store<AppState>, private route: ActivatedRoute) {
+  private open$: Observable<boolean>;
+
+  public constructor(private router: Router, private store: Store<AppState>) {
     if (environment.videoKey) {
       this.store.dispatch(new VideosAction.LoadVideos({videos: getAllVideos(), apiKey: environment.videoKey}));
     }
+
+    this.open$ = this.bindOpen();
```
kubedo8: Do we need this property still there?

kubedo8:
Did you try to create some projects? If not, please try to create project at least 3 times to ensure everything works properly.

src/app/dialog/dialog.service.ts
livthomas: No, but we might forget about these problems in a few months. I added it here in order to allow us to access dialog state in the future without implementing nasty workarounds as we did in the past.

kubedo8: Ok

livthomas:
You are right. There are still some problems. Do not merge this yet!

PR #886, opened 2019-06-26
Change select-item dropdown to cdk overlay #LMR-1842

kubedo8:


src/app/shared/select/select-item/select-item-dropdown/select-item-dropdown.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  @Input()
+  public selectedId: any;
+
+  @Input()
+  public fitParent: boolean;
+
+  @Output()
+  public selectItem = new EventEmitter<SelectItemModel>();
+
+  @ViewChild('selectItemDropdown', {static: true})
```
livthomas: You can use `static: false` here since it is not accessed in `ngOnInit`. It will be [the default option in Angular 9](https://angular.io/guide/static-query-migration).

PR #891, opened 2019-07-04
Use constraint icons in attribute type dialog #LMR-1870

livthomas:


livthomas:
Screenshot from 2019-07-04 15-36-44

PR #905, opened 2019-07-16
Cosmetic updates in UI

marvec:


package-lock.json
```diff
@@ -143,7 +143,7 @@
       "dependencies": {
         "source-map": {
           "version": "0.5.6",
-          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.6.tgz",
+          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.5.6.tgz",
```
livthomas: @marvec You should probably use newer version of NPM to avoid such changes.

PR #912, opened 2019-07-18
Create duration constraint #LMR-1587

kubedo8:


src/app/core/service/translation.service.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import {Injectable} from '@angular/core';
+import {I18n} from '@ngx-translate/i18n-polyfill';
+import {DurationUnit} from '../model/data/constraint-config';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class TranslationService {
```
livthomas: Why is this called `TranslationService` if it only provides methods specific for duration constraint?

src/app/dialog/attribute-type/form/constraint-config/duration/conversion/duration-constraint-config-form-conversions.component.html
```diff
@@ -0,0 +1,44 @@
...
+                   [formControlName]="conversionsControls.Value"
+                   [readOnly]="!editable || !(control.value.unit | durationConstraintUnitEditable)"
+                   min="1"
+                   [max]="control.value.unit | durationConstraintUnitMaxValue">
+
+          </td>
+          <td class="align-middle text-nowrap p-1">
+            <div class="info-column text-muted"
+                 i18n="@@constraint.duration.unit.conversion">
+              {control.value.unit, select, w {days} d {hours} h {minutes} m {seconds} s {milliseconds} }
```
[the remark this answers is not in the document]
kubedo8: Yes you are right. Days in week, hours in day etc..

src/app/core/service/translation.service.ts
kubedo8: Because it is prepared for the future. But I can remove it to something else..

PR #919, opened 2019-07-24
Add support for multiple query stem paths in Pivot, Gantt & Calendar #LMR-1911

kubedo8:


src/app/view/perspectives/gantt-chart/tasks/visualization/gantt-chart-visualization.component.ts
```diff
@@ -45,13 +45,19 @@ export class GanttChartVisualizationComponent implements OnChanges {
...
-    collectionConfigId: string;
+    resourceType: AttributesResourceType;
     changes: {attributeId: string; value: any}[];
   }>();
 
+  @Output()
+  public progressChange = new EventEmitter<{
+    dataResourceId: string;
+    resourceType: AttributesResourceType;
+    change: {attributeId: string; value: any};
```
livthomas: I can see this data structure defined multiple times. Maybe it would be better to extract it as an interface.

src/i18n/messages.cs.xlf
```diff
@@ -569,7 +569,7 @@
         <source>
     Example: 1w2d5h20m30s
   </source><target>
-    Příklad: 1w2d5h20m30s
+    Příklad: 1t2d5h20m30s
```
livthomas: I just wonder if it is not better to render these examples dynamically.

src/app/dialog/dialog-routing.module.ts
```diff
@@ -105,7 +105,7 @@ const routes: Routes = [
     outlet: 'dialog',
   },
   {
-    path: `${DialogPath.CALENDAR_EVENT}/:calendarId/:time/:documentId`,
+    path: `${DialogPath.CALENDAR_EVENT}/:calendarId/:time/:documentId/:stemIndex`,
```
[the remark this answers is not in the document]
kubedo8: After new changes, we can have same query stems in query. In this case I need to know which config I should use to display and edit data in dialog, because `collectionId` is not enough.

src/i18n/messages.cs.xlf
kubedo8: Yes I was thinking about that.. I will do it

src/app/core/store/calendars/calendar-config-converter.ts
```diff
@@ -0,0 +1,66 @@
...
+    default:
+      return convertCalendarDtoToModelV0(config);
+  }
+}
+
+function convertCalendarDtoToModelV1(config: any): CalendarConfig {
+  return config;
+}
+
+function convertCalendarDtoToModelV0(config: any): CalendarConfig {
```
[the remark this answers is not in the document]
kubedo8: That's good idea.

PR #920, opened 2019-07-24
Add file attachment constraint #LMR-1898

livthomas:


src/app/core/rest/file-attachment-api.service.ts
```diff
@@ -0,0 +1,110 @@
...
+      `${this.filesUrl(path)}/collection/${path.collectionId}/${path.documentId}/${path.attributeId}/details`
+    );
+  }
+
+  public getFilesByLinkType(path: FileApiPath): Observable<FileAttachmentDto[]> {
+    return this.http.get<FileAttachmentDto[]>(`${this.filesUrl(path)}/link/${path.collectionId}`);
+  }
+
+  public getFilesByLinkInstance(path: FileApiPath): Observable<FileAttachmentDto[]> {
+    return this.http.get<FileAttachmentDto[]>(`${this.filesUrl(path)}/link/${path.collectionId}/${path.documentId}`);
```
kubedo8: I think we should rename these parameters to `resourceId` and `dataId` or use another object for links, because it can be very confusingl

src/app/shared/data-input/files/button/file-attachment-button.component.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private onRefreshUrlSuccess(fileAttachments: FileAttachment[]) {
+    const fileAttachment = fileAttachments.find(file => file.id === this.fileAttachment.id);
+    this.downloadFileAttachment(fileAttachment);
+  }
+
+  private onRefreshUrlFailure() {
+    this.downloading$.next(false);
+    // TODO notification
```
kubedo8: When these TODO's will be implemented?

livthomas: Yeah, I forgot about this one.

src/app/core/rest/file-attachment-api.service.ts
livthomas: I have modified `FileApiPath` to be like `FileAttachment`. But `FileAttachmentDto` needs to follow backend structure.

PR #923, opened 2019-07-25
Limit precision in coordinates config #LMR-1860

livthomas:
Screenshot from 2019-07-25 17-56-42
Screenshot from 2019-07-25 17-56-57

### DIFF
--- a/src/app/dialog/attribute-type/form/constraint-config/coordinates/coordinates-constraint-config-form.component.html
+++ b/src/app/dialog/attribute-type/form/constraint-config/coordinates/coordinates-constraint-config-form.component.html
@@ -24,35 +24,28 @@
   <div class="form-group">
     <label [for]="controls.Precision"
            i18n="@@constraint.coordinates.precision.label">
-      Fractional digits
+      Precision
     </label>
-    <input [id]="controls.Precision"
-           [formControlName]="controls.Precision"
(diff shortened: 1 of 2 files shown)


src/app/dialog/attribute-type/form/constraint-config/coordinates/coordinates-constraint-config-form.component.html
```diff
@@ -24,35 +24,58 @@
...
-      You need to enter a number lesser than or equal to {{precisionControl.errors?.max?.max}}.
-    </div>
+    <select [id]="controls.Precision"
+            [formControlName]="controls.Precision"
+            class="custom-select"
+            data-test="coordinates-constraint-precision-select">
+      <ng-container *ngIf="formatControl.value === coordinatesFormat.DecimalDegrees; else dmsPrecisions">
+        <option i18n="@@constraint.coordinates.precision.dd.0"
+                value="0">
+          0 digits (~ 100 km)
```
kubedo8: Wouldn't it be better to create numeric enums or constants and iterate over it? Also translation will be quite easier.

livthomas: Thanks. I have changed that.

PR #942, opened 2019-08-05
Add support for links in kanban perspective #LMR-1919

kubedo8:


src/app/shared/utils/data.utils.ts
```diff
@@ -632,3 +634,14 @@ export function formatSelectDataValue(value: any, config: SelectConstraintConfig
 export function isSelectDataValueValid(value: any, config: SelectConstraintConfig): boolean {
   return config && config.options.some(option => String(option.value) === String(value));
 }
+
+function getSelectSaveValue(value: any, config: SelectConstraintConfig): any {
+  if (config.displayValues) {
+    const displayOption = config.options.find(option => option.displayValue === value);
+    if (displayOption) {
+      return displayOption.value;
```
livthomas: Nice catch! This needs to be handled in the new data value objects as well.

src/app/view/perspectives/kanban/columns/column/kanban-column.component.ts
```diff
@@ -75,9 +94,15 @@ export class KanbanColumnComponent implements OnInit, OnChanges {
...
   public collections: Collection[];
 
+  @Input()
+  public linkTypes: LinkType[];
+
   @Input()
   public documents: DocumentModel[];
 
+  @Input()
+  public linkInstances: LinkInstance[];
```
livthomas: Why do we need to have instances of all these entities here if we already have an abstract structure such as `KanbanCard` which already carries those entities?

src/app/view/perspectives/kanban/kanban-perspective.module.ts
```diff
@@ -27,43 +27,53 @@ import {SharedModule} from '../../../shared/shared.module';
...
 import {KanbanPostItDraggablePipe} from './pipes/kanban-post-it-draggable.pipe';
-import {BsDropdownModule} from 'ngx-bootstrap';
-import {FilterWritableCollectionsPipe} from './pipes/filter-writable-collections.pipe';
+import {FilterWritableResourcesPipe} from './pipes/filter-writable-resources.pipe';
 import {PickerModule} from '../../../shared/picker/picker.module';
 import {KanbanSelectedItemPipe} from './pipes/kanban-selected-item.pipe';
+import {KanbanRestrictedSelectItemIdsPipe} from './pipes/kanban-restricted-select-item-ids.pipe';
+import {DropdownModule} from '../../../shared/dropdown/dropdown.module';
+import {ChooseLinkDocumentModalComponent} from './modal/choose-link-document/choose-link-document-modal.component';
+import {ModalModule as NgxModalModule} from 'ngx-bootstrap';
```
livthomas: We should avoid importing the whole `ngx-bootstrap` and should use something like this instead:
```typescript
import { ModalModule as NgxModalModule } from 'ngx-bootstrap/modal';
```

src/app/view/perspectives/map/map-perspective.component.html
```diff
@@ -1,4 +1,4 @@
-<ng-container *ngIf="validQuery$ | async; else invalidQuery">
+<ng-container *ngIf="query$ | async | perspectiveDisplayable:1; else invalidQuery">
```
livthomas: Nice improvement!

src/app/core/store/documents/documents.effects.ts
```diff
@@ -107,6 +110,32 @@ export class DocumentsEffects {
...
+    mergeMap(action => {
+      const {document, otherDocumentId, linkTypeId, callback} = action.payload;
+      const documentDto = convertDocumentModelToDto(document);
+
+      return this.documentService.createDocument(documentDto).pipe(
+        map(dto => convertDocumentDtoToModel(dto)),
+        mergeMap(newDocument => {
+          const linkInstance: LinkInstance = {documentIds: [newDocument.id, otherDocumentId], linkTypeId};
+          const linkInstanceDto = convertLinkInstanceModelToDto(linkInstance);
+          return this.linkInstanceService.createLinkInstance(linkInstanceDto).pipe(
```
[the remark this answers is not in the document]
kubedo8: Situation is quite different here. I need to scroll to new post-it after document and link is created. The other solution is to use callback after document creation, execute link action with callback and then scroll the post-it. But I'm afraid it will not work. Callback in these effects are executed before object is added to store, so post-it is not created yet. 

Solution 1: Use existing code.
Solution 2: Execute callbacks in effects after objects are added to store.
Solution 3: ??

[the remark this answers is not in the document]
kubedo8: So what do you recommend?

src/app/view/perspectives/kanban/columns/column/kanban-column.component.ts
kubedo8: It's quite same reason as above. I need to execute store action in this component, because it's necessary to scroll down to new post-it after successful action. These data are needed to correctly create document. The other solution would be to emit action to parent component and then call function on `ViewChild` from parent.

src/app/core/store/documents/documents.effects.ts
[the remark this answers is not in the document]
livthomas: I can see that the callbacks for both document and link instance creation are not implemented the best way and they are executed before the success action. We should probably change that, however, it might break something so it does not make sense to do that in this PR. I think your solution is good for now. But we should try to avoid creating custom actions and action composition whenever possible.

PR #948, opened 2019-08-08
Switch building completely to CircleCI #LMR-1964

livthomas:


livthomas:
Because of the CircleCI memory limits (4GB), we cannot use it to run E2E tests at the moment.

PR #949, opened 2019-08-08
Add new user invite button #LMR-1894

marvec:


src/app/core/model/invitation-type.ts
```diff
@@ -0,0 +1,34 @@
...
+import {ResourceType} from './resource-type';
+
+export enum InvitationType {
+  JoinOnly = 'JOIN_ONLY',
+  ReadOnly = 'READ_ONLY',
+  ReadWrite = 'READ_WRITE',
+  Manage = 'MANAGE',
+}
+
+export const invitationTypesMap: {[id: string]: InvitationType} = {
```
kubedo8: We use `Record<string, InvitationType>` syntax.

src/app/shared/top-panel/user-panel/invite-user/invite-user-dialog/invite-user-dialog.component.html
```diff
@@ -0,0 +1,117 @@
...
+  *ngIf="{
+      isOrganizationManager: (organization$ | async) | permissions:organizationType:'manage' | async,
+      isProjectManager: (project$ | async) | permissions:projectType:'manage' | async
+    } as managers">
+
+  <span class="title" i18n="@@inviteUser.dialog.title">Invite colleagues</span>
+
+  <div class="body p-3">
+
+    <div class="d-block">
```
kubedo8: This component is too long :( It would be better to extract it.

PR #957, opened 2019-08-16
Align scrolling behavior across perspectives #LMR-1942

livthomas:


livthomas:
If you mean the event listeners then probably no. I have tried it several times before but the event is not emitted when you listen directly to the DOM events. It's probably some special jQuery thing.

kubedo8:
I mean something like this https://valor-software.com/ngx-bootstrap/#/alerts

livthomas:
Ok, I'll try to switch it to `ngx-bootstrap` to avoid using jQuery.

livthomas:
@kubedo8 Done.

PR #981, opened 2019-09-17
FIRST MERGE BE: Fix blockly serial execution and link value update detection, add new blocks

marvec:
… blocks for seqences, user, colors


src/app/core/store/sequences/sequences.state.ts
```diff
@@ -0,0 +1,51 @@
...
+import {createEntityAdapter, EntityState} from '@ngrx/entity';
+import {AppState} from '../app.state';
+import {createSelector} from '@ngrx/store';
+import {Sequence} from '../../model/sequence';
+
+export interface SequencesState extends EntityState<Sequence> {
+  loaded: boolean;
+}
+
+export const SequencesAdapter = createEntityAdapter<Sequence>({
```
kubedo8: constants and variable names start with lower case character

src/app/shared/sequences/sequence-item/sequence-item.component.scss
```diff
@@ -0,0 +1,3 @@
+.sequence-name {
+  vertical-align: middle;
```
kubedo8: for such cases you can use `align-middle` class

PR #1001, opened 2019-10-08
Refactor constraint interfaces #LMR-1926

livthomas:


src/app/core/service/translation.service.ts
```diff
@@ -34,7 +34,7 @@ export class TranslationService {
   public translateDurationUnit(unit: DurationUnit): string {
     return this.i18n(
       {
-        id: 'constraint.duration.unit',
+        id: 'index.ts.duration.unit',
```
kubedo8: Just a refactor typo

src/app/shared/pipes/constraint-type-icon-title.pipe.ts
```diff
@@ -34,7 +34,7 @@ export class ConstraintTypeIconTitlePipe implements PipeTransform {
 
     return this.i18n(
       {
-        id: 'constraint.type.icon.title',
+        id: 'index.ts.type.icon.title',
```
kubedo8: Another refactor typo

kubedo8:
Looks very good! I see how it simplify our work with constraints.
I will test it in few hours.

marvec:
Did you test the constraints in table? I cannot start entering values for percentage and duration constraint properly. Percentage use wrong formatting and duration always enters 0. Also, kanban-converter.ts seems to contain errors.

livthomas:
I've resolved all the issues you pointed out except for the one in duration data input. The behavior is exactly the same as in our production instance so if we want to change that it will probably be a task for @kubedo8 as he's going to optimize `DurationDataValue` anyway. It is also not possible to enter zero into duration data input at the moment (and it seems it's always been like this).

livthomas:
I've fixed entering numbers to duration data input.

livthomas:
The resizing of data inputs should work now.

PR #1037, opened 2019-10-26
React to realod push notification

marvec:


src/app/core/pusher/pusher.service.ts
```diff
@@ -330,7 +330,25 @@ export class PusherService implements OnDestroy {
       if (this.isCurrentWorkspace(data)) {
         this.store$.dispatch(
           new DocumentsAction.Get({
-            query: {stems: [{collectionId: data.id}], page: 0, pageSize: 100},
+            query: {stems: [{collectionId: data.id}], page: 0, pageSize: 200},
+          })
+        );
+      }
+    });
+    this.channel.bind('Collection:reload', data => {
```
kubedo8: Why we send push notification with whole object? I think it's not needed in this case.

src/app/core/pusher/pusher.service.ts
```diff
@@ -321,7 +321,7 @@ export class PusherService implements OnDestroy {
       if (this.isCurrentWorkspace(data)) {
         this.store$.dispatch(
           new DocumentsAction.Get({
-            query: {stems: [{collectionId: data.object.id}], page: 0, pageSize: 100},
+            query: {stems: [{collectionId: data.object.id}], page: 0, pageSize: 200},
```
kubedo8: 200 is new magic number?

src/app/core/pusher/pusher.service.ts
```diff
@@ -330,7 +330,25 @@ export class PusherService implements OnDestroy {
...
+          })
+        );
+      }
+    });
+    this.channel.bind('Collection:reload', data => {
+      if (this.isCurrentWorkspace(data)) {
+        this.store$.dispatch(
+          new DocumentsAction.Get({
+            query: {stems: [{collectionId: data.object.id}], page: 0, pageSize: 200},
+          })
```
kubedo8: This action has filter inside for already loaded queries (look at its effect). I think we need to force refresh in this case.

src/app/core/pusher/pusher.service.ts
marvec: 100 was too little. It did not read the whole collection. I would even make it 500. The problem is that paging does not work properly. Remaining data are never read. WDYT?

src/app/core/pusher/pusher.service.ts
marvec: How do I do that?

src/app/core/pusher/pusher.service.ts
marvec: I used what was already there not to have to create way too much code. Do you think it is worth fixing?

src/app/core/pusher/pusher.service.ts
kubedo8: I think we should implement paging or read the whole collection.

src/app/core/pusher/pusher.service.ts
kubedo8: You can pass boolean value to action and use it in effect.

src/app/core/pusher/pusher.service.ts
kubedo8: Maybe it's not worth it.

PR #1063, opened 2019-11-09
Rich text editor

marvec:


src/app/shared/data-input/data-input.component.ts
```diff
@@ -64,7 +65,7 @@ export class DataInputComponent implements OnChanges, OnDestroy {
   public value: any;
 
   @Input()
-  public skipValidation = false;
+  public configuration: DataInputConfiguration = {skipValidation: false, fromQuery: false};
```
kubedo8: You can set both variables in interface as optional and then it won't be necessary to set initial value.

src/app/shared/data-input/data-input.component.html
```diff
@@ -3,7 +3,7 @@
   <address-data-input *ngSwitchCase="constraintType.Address"
                       [focus]="focus"
                       [readonly]="readonly"
-                      [skipValidation]="skipValidation"
+                      [skipValidation]="configuration.skipValidation"
```
kubedo8: I think you can pass new interface everywhere. It will be necessary in future change.

src/app/shared/data-input/data-input.component.ts
```diff
@@ -64,7 +65,7 @@ export class DataInputComponent implements OnChanges, OnDestroy {
...
 
   @Input()
   public resizeToContent = false;
```
kubedo8: Maybe we can add this attribute to configuration too.

src/app/shared/modal/text-editor/text-editor-modal.component.ts
```diff
@@ -0,0 +1,170 @@
...
+      newValid = filterText.length >= this.minLength;
+    }
+
+    if (this.maxLength) {
+      newValid = newValid && filterText.length <= this.maxLength;
+    }
+
+    if (newValid !== this.valid) {
+      this.valid = newValid;
+      setTimeout(() => this.editoreHeight());
```
kubedo8: Just a small typo

src/app/shared/data-input/data-input.component.html
```diff
@@ -132,6 +132,20 @@
                    (dataBlur)="onDataBlur()">
   </user-data-input>
 
+  <ng-container *ngSwitchCase="constraintType.Text">
+    <rich-text-data-input *ngIf="!configuration.fromQuery; else textDataInput"
```
kubedo8: How is html handled in query?

src/app/shared/data-input/data-input.component.html
marvec: I don't know, the other properties do not make any sense for the component and it would create a fake impression that they make a difference or influence something...

src/app/shared/data-input/data-input.component.ts
marvec: They are optional. I did this just to be sure and careful that the previous behaviour is not changed.

src/app/shared/data-input/data-input.component.html
marvec: This requires a later change on backend.

PR #1095, opened 2019-11-28
Share dialog redesign #LMR-2113

marvec:


src/app/shared/modal/share-view/body/share-view-dialog-body.component.scss
```diff
@@ -1,3 +1,7 @@
 table {
   table-layout: fixed;
 }
+
+.tools {
+  width: 8.75rem;
```
kubedo8: Are you sure about that? Dialog can have various width, and especially on mobile devices this value may be very big.

src/app/shared/pipes/role-human-readable.pipe.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public transform(roles: string[]): string {
+    let roleText = 'UNKNOWN';
+
+    if (roles.findIndex(role => role === 'MANAGE') >= 0) {
+      roleText = 'MANAGE';
+    } else if (roles.findIndex(role => role === 'WRITE') >= 0) {
+      roleText = 'WRITE';
+    } else if (roles.findIndex(role => role === 'READ') >= 0) {
+      roleText = 'READ';
```
kubedo8: Maybe you can use `Role` constants

PR #1213, opened 2020-03-12
Upgrade to Angular9 #LMR-2227

kubedo8:


env-vars.js
```diff
@@ -29,7 +29,7 @@ config.VIDEO_KEY = env.VIDEO_KEY;
 
 writeFileSync('./src/environments/.env.json', JSON.stringify(config));
 
-const buildSwitches = `--aot --base-href=${config.PUBLIC_PATH} ${
+const buildSwitches = `${config.LUMEER_ENV === 'production' ? '--prod' : '--aot'} --base-href=${config.PUBLIC_PATH} ${
```
marvec: Isn' the --aot flag deprecated/non-functional at all?

livthomas: It's not. It's just that the AOT should be turned on by default. The update script from Angular CLI puts it to `angular.json` but I can see it hasn't been run here.

PR #1251, opened 2020-04-02
Fix moment parsing and table cell group height checking

marvec:


src/app/view/perspectives/table/header/column-group/table-column-group.component.ts
```diff
@@ -89,9 +96,33 @@ export class TableColumnGroupComponent implements OnChanges, AfterViewChecked {
     const height = element.offsetHeight;
 
     if (height) {
-      const tableElement = getTableElement(this.cursor.tableId);
-      tableElement?.style.setProperty('--column-group-height', `${height}px`);
+      this.checkElementHeight(this.cursor.tableId, height);
+      //const tableElement = getTableElement(this.cursor.tableId);
+      //tableElement?.style.setProperty('--column-group-height', `${height}px`);
```
kubedo8, line 101: I think we can remove these comments

marvec: sure thing!

PR #1551, opened 2020-12-02
Fix links table width

marvec:


src/app/shared/links/links-list/links-accordeon.component.html
```diff
@@ -19,6 +19,7 @@
           <links-list-table
             *ngIf="{otherCollection: linkType | linkTypeOtherCollection:(collections$ | async):collection.id} as linkData"
             class="d-block mt-2"
+            [visible]="data.openedGroups.get(linkType.id)"
```
kubedo8, line 22: Why did you use `Map` ? We use `Record` or simple `{}` .. calling methods in template is not recommended.

PR #1702, opened 2021-03-21
Dynamic environments

kubedo8:


src/app/app.module.ts
```diff
@@ -63,18 +64,26 @@ export const angularticsSettings: Partial<Angulartics2Settings> = {
     Angulartics2Module.forRoot(angularticsSettings),
   ],
   providers: [
+    {
+      provide: APP_INITIALIZER,
+      useFactory: (cs: ConfigurationService) => () => cs.loadConfiguration(),
```
marvec, line 69: I would suggest to use a different property name than cs because it resembles the CS locale. configServ at least or so...

kubedo8: It is used only here, but whatever.. I will change it

PR #1736, opened 2021-04-15
Add the ability to clean sample project data

marvec:


kubedo8:
Looks like the main code is missing here.

PR #1867, opened 2021-12-19
Get variable and send SMTP blocks

marvec:


src/app/shared/modal/attribute-function/attribute-function-modal.component.ts
```diff
@@ -94,6 +98,17 @@ export class AttributeFunctionModalComponent implements OnInit {
...
+      select(selectProjectByWorkspace),
+      switchMap(project => {
+        return this.store$.pipe(
+          select(selectResourceVariablesByResourceType(project.id, ResourceType.Project)),
+          map(resourceVariables => {
+            return resourceVariables.map(variable => variable.key);
+          })
+        );
+      })
+    );
```
kubedo8, line 111: We should create special selectors when some code is copied 3 times.

